PR #1, opened 2019-04-20
add implementation

amesgen:
This PR adds the implementation of the client (see http4s/http4s#2490). It "needs" http4s/http4s#2521 and the http4s client test module being published in order to remove the two TODOs. Both changes should be binary compatible (?).

The dependency to `reactive-streams-flow-adapters` can be removed as soon as a new version of `reactive-streams` is published, as it will take advantage of multi-release JARs (see reactive-streams/reactive-streams-jvm#430).


core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,136 @@
...
+          }.toList),
+          httpVersion = res.version match {
+            case HttpClient.Version.HTTP_1_1 => HttpVersion.`HTTP/1.1`
+            case HttpClient.Version.HTTP_2 => HttpVersion.`HTTP/2.0`
+          },
+          body = FlowAdapters
+            .toPublisher(res.body)
+            .toStream[F]
+            .flatMap(bs =>
+              Stream.fromIterator(bs.asScala.map(Chunk.byteBuffer).iterator).flatMap(Stream.chunk))
```
amesgen: This could also be written as `Stream(bs.asScala.map(Chunk.byteBuffer): _*)`, but it won't support 2.13.

PR #2, opened 2019-04-26
update to http4s 0.20

amesgen:
Now using http4s/http4s#2521.


ChristopherDavenport:
Ross! The outrage!

rossabaker:
Sorry, I was trying to get the site revamped last weekend and then ran out of weekend.  I'll turn that off until I can finish it.

PR #13, opened 2019-05-06
Remove scalajs and crossproject plugins

rossabaker:
Not sure whether scala.js even supports this client, but http4s doesn't.  If the dependencies are ever there, we can readd it.  It's just noise right now.

Replaces #9.


rossabaker:
Rebased on master to see if xenial fixes the openjdk12 problem.

PR #4, opened 2019-05-06
Support 2.13.0-M5

rossabaker:
These are the versions of Scala supported by http4s-0.20.

We'll go 2.13.0-RC1 as soon as http4s can, but we should have shipped this days ago, and I'll bet RC1 support in http4s is still a couple weeks away at least.

PR #15, opened 2019-05-06
Swap out sbt-microsite for paradox

rossabaker:
Gains:
1. No jekyll, no hugo, no docker. Only Scala.
2. Search!
3. (subjective) *way* more attractive

Losses:
1. github4s integration (sbt-ghpages should still work)
2. Instutional knowledge of jekyll and hugo


rossabaker:
There's more to do after #4, but one step at a time.

PR #17, opened 2019-05-18
Enforce scalafmt

rossabaker:
* Uses the same settings as http4s/http4s repo.
* Adds a `validate` alias, run by CI


.travis.yml
```diff
@@ -20,9 +20,7 @@ install:
   - gem install jekyll -v 3.2.1
 
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION test
-  - sbt ++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues
-  - sbt ++$TRAVIS_SCALA_VERSION docs/makeSite
+  - sbt ++$TRAVIS_SCALA_VERSION validate
```
rossabaker: This use of `validate` neglects that we only build the docs on one version.  We'll need to rethink this when the merge conflict from #4 hits us.

PR #22, opened 2019-05-23
Fix broken Scaladoc references

rossabaker:
[Oops](https://travis-ci.org/http4s/http4s-jdk-http-client/jobs/536301055#L245).


ChristopherDavenport:
If I had a nickel for every time I had broken builds that I only discovered on release this way.

I should really add `doc` to my ci.

rossabaker:
I thought makeSite did it, but we're not publishing the scaladoc on the site right now.

PR #24, opened 2019-05-23
Import the gpg key

rossabaker:
Depends on PGP_SECRET and PGP_PASSPHRASE, which I've already set up in Travis CI.

For posterity's sake: using key 3C477ED52BBC57DDE7D1FC8A14E4D722E40D4E4F


rossabaker:
Note that we're not actually testing this until the master build.

PR #28, opened 2019-05-25
Overhaul release process

rossabaker:
The good:

* Uses git versioning.  The version number is set by a tag.  If it is not a tagged build, it's set to the base version, plus the SHA, plus `-SNAPSHOT`.  For instance, this version is `0.1.0-3e20680a711b8844c7d51d0ecfdcb1261472f89d-SNAPSHOT`.  If I tagged and pushed `v0.1.0`, the resulting build would be version `0.1.0`.

* Adds a `release` stage, executed only on tags.  This is identical to the `publish` stage, but wraps the command in `sonatypeOpen` and `sonatypeRelease`.  This is an improvement on `sonatypeReleaseAll`, which interferes with other projects in the organization being released at the same time.

* Eliminates sbt-release.  Special behavior is now driven by tags, and Travis CI no longer commits to git except to publish the site.  This is a source of failed release builds when changes are happening quickly.

The bad:

* We have to bump the baseVersion to get sensible snapshot definitions.

* Nothing checks tags for snapshot dependencies, though we could scavenge this code.

* Nothing checks that the tag matches the base version, though we could add that if we're afraid of typos in a tag.

* There's no pull request that signifies a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ java12: &java_12 openjdk12
 stages:
   - name: tests
   - name: publish
-    if: (branch = master AND type = push)
+    if: (branch = master AND type = push AND tag IS blank)
+  - name: release
+    if: tag IS present
 
 jobs:
   include:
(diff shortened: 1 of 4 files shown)


rossabaker:
An alternative would be to keep sbt-release.  This would mitigate the aforementioned negatives.  It does not solve the problem of detecting whether to execute `+publish` or `release with-defaults`.  We could create conditional release steps and always use sbt-release, opting out of the sonatype steps when it's a snapshot.  I have very bad experience with `sonatypeOpen` and sbt-release, but maybe it's better now.

amesgen:
We basically use a setup like this now 👏🏿

PR #53, opened 2019-07-31
intial WebSocket support

amesgen:


core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWebSocketClient.scala
```diff
@@ -0,0 +1,129 @@
...
+            } yield (webSocket, queue, sendSem)
+          } {
+            case (webSocket, _, _) =>
+              for {
+                isOutputOpen <- F.delay(!webSocket.isOutputClosed)
+                closeOutput = fromCompletableFuture(
+                  F.delay(webSocket.sendClose(JWebSocket.NORMAL_CLOSURE, ""))
+                )
+                _ <- closeOutput.whenA(isOutputOpen)
+                _ <- F.delay(webSocket.abort())
```
amesgen: I am not sure if closing both sides apruptly is a good idea immediately after sending a close frame, but it won't matter a lot and it protects us from a server which continues to send frames afterwards. Maybe sleep for a second beforehand?

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
```diff
@@ -0,0 +1,167 @@
...
+                  case _ => F.unit
+                }
+                .collect { case wsdf: WSDataFrame => wsdf }
+                .through(groupFrames[F])
+            override def subprocotol: Option[String] = conn.subprotocol
+            override def closeFrame: TryableDeferred[F, WSFrame.Close] = recvCloseFrame
+          }
+    }
+
+  private[jdkhttpclient] def groupFrames[F[_]]: Pipe[F, WSDataFrame, WSDataFrame] = {
```
amesgen: I could not come up with a straightforward way to write this with the existing combinators...

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWebSocketClient.scala
```diff
@@ -0,0 +1,129 @@
...
+                  webSocket.getSubprotocol.some.filter(_.nonEmpty)
+              }
+          }
+    }
+
+  /** A `WebSocketClient` wrapping the default `HttpClient`. */
+  def simple[F[_]](implicit F: ConcurrentEffect[F]): F[WebSocketClient[F]] =
+    F.delay(HttpClient.newHttpClient()).map(apply(_))
+
+  private def toCompletionStage[F[_], A](fa: F[A])(implicit F: Effect[F]): CompletionStage[A] = {
```
amesgen: same story as for `fromCompletableFuture` - should this be an internal helper function in `http4s-core`?

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWebSocketClient.scala
rossabaker: I don't know the web socket protocol very well, but [section 7.1.1 of RFC6455 says](https://tools.ietf.org/html/rfc6455#page-41):

> The underlying TCP connection, in most normal cases, SHOULD be closed
   first by the server, so that it holds the TIME_WAIT state and not the
   client (as this would prevent it from re-opening the connection for 2
   maximum segment lifetimes (2MSL), while there is no corresponding
   server impact as a TIME_WAIT connection is immediately reopened upon
   a new SYN with a higher seq number).  In abnormal cases (such as not
   having received a TCP Close from the server after a reasonable amount
   of time) a client MAY initiate the TCP Close.  As such, when a server
   is instructed to _Close the WebSocket Connection_ it SHOULD initiate
   a TCP Close immediately, and when a client is instructed to do the
   same, it SHOULD wait for a TCP Close from the server.

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWebSocketClient.scala
rossabaker: I want it in cats-effect, but I've been shot down.  Yes, http4s-core internals seems like an appropriate place, so other modules can use it.

rossabaker: And probably call it unsafeToCompletionStage.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+  * A websocket request.
+  *
+  * @param uri The URI.
+  * @param headers The headers to send. Some websocket clients reject protocol-specific headers.
+  * @param method The method of the intial HTTP request. Ignored by some clients.
+  * @param subprotocols The subprotocols, descending from most prefered.
+  */
+case class WSRequest(
```
rossabaker: Are we reasonably certain no other fields will be added to this?  The builder pattern instead of a case class will promote binary compatibility if this is open to future extension.

rossabaker: Is there no HTTP version in these requests?

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
```diff
@@ -0,0 +1,167 @@
...
+    uri: Uri,
+    headers: Headers = Headers.empty,
+    method: Method = Method.GET,
+    subprotocols: List[String] = List.empty
```
rossabaker: I would bet not all `String`s are valid subprotocols.  There might be a missing type here.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
```diff
@@ -0,0 +1,167 @@
...
+)
+
+sealed trait WSFrame
```
rossabaker: Do these need to be different from the server model?  It would seem nice to unify.

rossabaker: `extends Product with Serializable` will help inference.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
```diff
@@ -0,0 +1,167 @@
...
+
+trait WSConnection[F[_]] {
+
+  /** Send a single websocket frame. The sending side of this connection has to be open. */
+  def send(wsf: WSFrame): F[Unit]
+
+  /** Send multiple websocket frames. Equivalent to multiple `send` calls, but at least as fast. */
+  def sendMany[G[_]: Traverse, A <: WSFrame](wsfs: G[A]): F[Unit]
+
+  /** A `Pipe` which sends websocket frames and emits `()`'s when they are sent. */
```
rossabaker: One `()` per frame or one per batch?

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  /** Send a single websocket frame. The sending side of this connection has to be open. */
+  def send(wsf: WSDataFrame): F[Unit]
+
+  /** Send multiple websocket frames. Equivalent to multiple `send` calls, but at least as fast. */
+  def sendMany[G[_]: Traverse, A <: WSDataFrame](wsfs: G[A]): F[Unit]
+  final def sendPipe: Pipe[F, WSDataFrame, Unit] = _.chunks.evalMap(sendMany(_))
+
+  /** Send a Ping frame. */
+  def sendPing(data: ByteVector = ByteVector.empty): F[Unit]
```
rossabaker: Does this (and `sendClose`) have a reasonable default implementation?

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
rossabaker: It would be nice if we got consistent between `WS` and `WebSocket` prefix.  This applies to core, as well.

docs/src/main/mdoc/index.md
```diff
@@ -95,16 +97,100 @@ def fetchStatusInefficiently[F[_]: ConcurrentEffect](uri: Uri): F[Status] =
...
 Clients created with this back end do not need to be shut down.
 
-## Further reading
+### Further reading
 
 For more details on the http4s-client, please see the [core client
-documentation][client].
+documentation][http4s-client].
+
+## WebSocket client
```
rossabaker: This might justify a second page in the docs.

docs/src/main/mdoc/index.md
```diff
@@ -95,16 +97,100 @@ def fetchStatusInefficiently[F[_]: ConcurrentEffect](uri: Uri): F[Status] =
...
+      // "faster" than individual `send` calls
+      _ <- conn.sendMany(List(
+        WSFrame.Text("is often"),
+        WSFrame.Text("disappointing.")
+      ))
+      received <- conn
+        // a backpressured stream of incoming frames
+        .receiveStream
+        // we do not care about Binary frames (and will never receive any)
+        .collect { case WSFrame.Text(str, _) => str }
```
rossabaker: I wonder if we're missing a whole class of web socket entity codecs. :thinking:

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
ChristopherDavenport: Isn't Websockets ONLY 1.1?

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWebSocketClient.scala
amesgen: Thx for the pointer - even though aborting immediately is allowed, it is discouraged, and closing it after a timeout is optional - I will remove the line.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: yes, see e.g. [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism)

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: Not in terms of another method, as `send` can only send data frames.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: actually, there are plans to support HTTP 2: [RFC](https://tools.ietf.org/html/draft-hirano-httpbis-websocket-over-http2-00)

amesgen: I changed it to `WS` (for now).

amesgen: This API will break when a common websocket api is in http4s(-core). IMHO I don't think bincompat dances are worth it right now.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: A better representation would express the subprotocols as a `ListSet`, and replace `String` by a newtype ensuring that it is a value HTTP header value (`field-name` from RFC 7230). I think this adds zero value in practice, as subprotocols are 1. rarely used and 2. almost always static.

If I think about it, it is probably better to remove this from  `WSRequest` and let the user supply a `Sec-WebSocket-Protocol` header on their own.

docs/src/main/mdoc/index.md
amesgen: What is a websocket entity codec about? Is there sth comparable in another ws client api?

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: I am confused - http4s `RequestCookie`s are not validated. Am I missing sth?

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: The existing model has some minor issues IMO, so I decided to try sth different in this draft.

 - exposes implementation details (continuation frames should not be part of the public API, you have to parse the close reason manually from a `Close`)
 - `WebSocketFrame` is not sealed
 - creation of a `Close` frame results in an either to prevent invalid close codes - IMHO such validations only result in one more `F.fromEither` and are therefore annoying. I would propose to shift this kind of validation to send time.

An advantage of the server model is that it allows to send text frames with binary data - but I would argue that this is rarely used in practice, and it is supported by neither the JDK ws client nor okhttp (but asynchttpclient allows this).

docs/src/main/mdoc/index.md
amesgen: What is a web socket entity codec? I don't think I understand what it is supposed to do - "websocket interaction" seems to be to broad to be put into a useful codec. Are there other WS client APIs with a similar concept?

amesgen:
Some comments by me are hidden because GitHub thinks the referenced code is outdated...

I have no clue why scalafmt fails on CI - `scalafmtCheckAll` runs fine on my machine...

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWebSocketClient.scala
rossabaker: I just revisited `fromFuture` last night, and am left reconsidering this: as `internals` are exposed in other repos in the org, and once we do that, we need to be careful about preserving binary compatibility.  This is one of the disadvantages to the polyrepo approach.  Maybe `org.http4s.util` is a legitimate package, though I almost always regret utils libraries.

Or maybe we just copy and paste it everywhere.  I don't know.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
rossabaker: You mean invalid `RequestCookie`s can be created? It might be something we just haven't tightened down from our Spray origins.

I think I agree with the header approach.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
rossabaker: Would the `Close` validation be less annoying if we had a macro validate literal values for the code and/or had adequate constants for the common ones?  Our principle is generally to only allow valid representations at construction, with some concessions to `unsafeFromXXX` methods.  But we do need to keep usability in mind, I agree.

It's difficult when the spec and a subset of implementations support a feature. What might we do if we tried to send such in this or okhttp?  Fail with the web socket equivalent of an UnsupportedOperationException?  If we don't do that, what if we add an even less capable backend later? 🤔

docs/src/main/mdoc/index.md
rossabaker: It probably doesn't make sense as anything other than a pipe, on deeper thought. I was wondering about aggregating text frames into lines or a stream of JSON objects, like jawn's async mode.  I'm not sure if there are common pipes that everyone writes or not.  I'm commenting on a spec that I don't have a lot of practical experience with, so feel free to tell me I'm being silly.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
amesgen: Your point concerning different backends is very true - and it also applies to `Close` validation. Quote from `sendClose` from the JDK WS client:

> Status codes 1002, 1003, 1006, 1007, 1009, 1010, 1012, 1013 and 1015 are illegal. Behaviour in respect to other status codes is implementation-specific.

I am unsure about the gains of "irrepresentable illegal states" in this case - IMHO the knowledge "this `send` call might fail for a lot of reasons, but it won't fail because the close code was invalid or the reason was longer than 123 bytes" is not very useful. There is also no parsing etc. involved, which is a difference to e.g. `Uri`.

Also Haskell's `websockets` package does neither validate their WS ADT ([link](https://hackage.haskell.org/package/websockets-0.12.5.3/docs/Network-WebSockets.html#t:Message)) nor provide a validated newtype for close codes ([link](https://hackage.haskell.org/package/websockets-0.12.5.3/docs/Network-WebSockets.html#v:sendCloseCode)) (just as a comparison, I don't want to make an appeal-to-haskthority argument).

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
rossabaker: Ah, right.  I was thinking solely in terms of `WSFrame`, and not the intermediate types.

core/src/main/scala/org/http4s/client/jdkhttpclient/WebSocketClient.scala
rossabaker: Some of those JDK WS restrictions make sense as reserved for backend use as opposed to letting an application set them.  I'm comfortable allowing them in the model (they mean something to the receiver!) and having them fail on send.  And once we allow that, you're right, we don't get much out of range validation.  The inconsistency with the rest of the model bothers me, but I find your argument pragmatic.

rossabaker:
We can just drop 2.11, because http4s dropped it on master.

I think this branch came before https://github.com/http4s/http4s-jdk-http-client/pull/52, which might want to reformat some code when testing the merge commit.  I suspect if you merge master and reformat, this problem will go away.

docs/src/main/mdoc/index.md
amesgen: As you say, parsing a JSON stream is possible in two LOC due to the awesome `jawn-fs2`. But such abstractions could ofc be built on top of the WS API in the future, if they are considered necessary for other use cases I don't see right now.

amesgen:
Can I do sth to move this forward? I personally don't rely on this PR being merged in any way, but it would be nice to get feedback from other people using this.

Concerning the tests: I don't understand Travis - it seems to both fail and pass, and in the failing job, one sub-job succeeds....

rossabaker:
Sorry, I've been all consumed chasing the AHC issues.  We're preparing an 0.21 milestone for http4s, and I would love to release a milestone with this and get it out there.

I think this is ready when we get to the bottom of the failing test.  It's passing for you locally?

amesgen:
Yes, everything works locally. The tests did pass in the past (e.g. `27c90e7`, the failure there was due to scalafmt).

rossabaker:
This has the look of one of those fun race conditions like we fight in core only on Travis.  If we rerun the job, does it flicker?

amesgen:
Am I able to restart the job? I highly suspect the test results to have good statistical properties...

rossabaker:
You should be able to. If you can't, let me know. I'd like to dig in and theorize why it flickers, but core has kept me busy today.

amesgen:
I can't seem to find the corresponding button - but I can see it in projects I am a member of. Weird, I can't think of a reason to disallow this...

rossabaker:
I just tweaked the permission,  but I'm not sure how long it will take Travis to catch up.

amesgen:
Ok, I have no idea how to debug these random test failures. The last commit apparentely didn't change anything, but I will stop adding garbage commits for now.

rossabaker:
The toString method on `Close` is not particularly helpful.

rossabaker:
Oops.  I misread the error tonight.  The `toString` is not helpful, but the real problem is a missing second `Close`.

amesgen:
Yeah, overwriting `toString` of `Close` in `http4s-core` would be a good idea (the underlying "problem" is that it stores the encoded bytes and not close code + reason - the "new" ADT does the latter).

rossabaker:
In hacking a bit on a branch in this repo. If I get something that works, I'll push to this branch.

amesgen:
I replaced Blaze by [Java-Websocket](https://github.com/TooTallNate/Java-WebSocket) (ugly code) - now all tests pass, even after restarting: [branch log](https://github.com/http4s/http4s-jdk-http-client/commits/websockets-java-server)

rossabaker:
Nice. I thought about trying Jetty for the same reason. I'm happy rolling with that for now.  I'd like to figure out whether the failing test is an issue with the client or server, so I'm going to keep poking at it, but I think your branch unblocks us.

rossabaker:
The blaze stage is reading the Close and sometimes shutting down before handling it.  It's a blaze-server bug.  I think your branch is the way forward.

amesgen:
Thx for digging into Blaze internals -  I will push a cleaned up version of Java-Websockets soon.

core/src/test/scala/org/http4s/client/jdkhttpclient/JdkWSClientSpec.scala
```diff
@@ -0,0 +1,183 @@
...
+            val req = WSRequest(uri"ws://localhost:8080")
+            val p = for {
+              _ <- webSocket.connect(req).use { conn =>
+                conn.send(WSFrame.Text("hi blaze"))
+              }
+              _ <- Timer[IO].sleep(1.seconds)
+              _ <- webSocket.connectHighLevel(req).use { conn =>
+                conn.send(WSFrame.Text("hey blaze"))
+              }
+              _ <- Timer[IO].sleep(1.seconds)
```
rossabaker: The `expect` that I did on my branch means we can be more optimistic and finish quicker.  I'll probably port that over in a follow up.  At this point, I'd like to get this landed and get people trying it because it works.

amesgen: Ah yes, I always forget that matchers are impure...

amesgen:
JFTR: "design issues" to be resolved:
 - server and client APIs differ (e.g. different WS frame ADTs are used, see [here](https://github.com/http4s/http4s-jdk-http-client/pull/53#discussion_r309888243))
 - future extensions to `WSRequest` (HTTP version or WS protocol extensions like permessage-deflate)

rossabaker:
It's late here, but I'll cut a release in the morning.

PR #64, opened 2019-08-17
Scala 2.12.9

sullis:


rossabaker:
Thanks. I'd like to merge this after we get metals support: https://github.com/scalameta/metals/issues/856

PR #119, opened 2019-12-01
Add a nix shell

rossabaker:
Like http4s/http4s#3005, but particularly useful here, where a newer JDK is required.


ChristopherDavenport:
What does this do for us nix novices?

rossabaker:
If I run `nix-shell` from this directory, I am dropped into a bash session that has an SBT that runs on JDK11, regardless of what sbt or java (if any) is globally installed on the system.

amesgen:
I don't use nix (apart from GHCJS experiments), do you still want to get this merged? @rossabaker

rossabaker:
I have a gitignored version.  I find it particularly useful on this project, because it's the only one I use on Java 11.  I'll refresh it with an updated version, but if it doesn't bother you, I'd appreciate having it around.

amesgen:
Actually, I have been converted to NixOS recently, and I would love to nixify this repo! Do you think the setup in https://github.com/typelevel/case-insensitive/pull/210 is still a playground, or can we adopt it also here (I can prepare a PR)?

rossabaker:
I think it would be great to add it.  It's particularly useful on this repo, which requires JDK 11, which is awkwardly between the 8 of Scala or the 17 that's current.

I'm experimenting on my personal config with workflows to automate flake updates, but as a baseline, what's merged on case-insensitive is good.  Note that I'm assuming flakes.

amesgen:
Superseded by #556

Issue #116, opened 2019-11-27
Java IOException


yilinwei:
I'm getting a Java `IOException` when the websocket is closed from the server and I presume that the client is sending a close message. I'm haven't read the RFC to ascertain the correct behaviour (whether this is a misbehaving websocket server, or it can happen).

```scala
java.io.IOException: closed output
        at java.net.http/jdk.internal.net.http.RawChannelTube.write(RawChannelTube.java:340)
        at java.net.http/jdk.internal.net.http.websocket.TransportImpl.write(TransportImpl.java:114)
        at java.net.http/jdk.internal.net.http.websocket.TransportImpl$SendTask.tryCompleteWrite(TransportImpl.java:597)
        at java.net.http/jdk.internal.net.http.websocket.TransportImpl$SendTask.run(TransportImpl.java:547)
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler$CompleteRestartableTask.run(SequentialScheduler.java:147)
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler$SchedulableTask.run(SequentialScheduler.java:198)
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler.runOrSchedule(SequentialScheduler.java:271)
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler.runOrSchedule(SequentialScheduler.java:224)
        at java.net.http/jdk.internal.net.http.websocket.TransportImpl.sendClose(TransportImpl.java:271)
        at java.net.http/jdk.internal.net.http.websocket.WebSocketImpl.sendClose0(WebSocketImpl.java:337)
        at java.net.http/jdk.internal.net.http.websocket.WebSocketImpl.sendClose(WebSocketImpl.java:307)
        at org.http4s.client.jdkhttpclient.JdkWSClient$.$anonfun$apply$17(JdkWSClient.scala:91)
        at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87)
        at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:355)
        at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:376)
        at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:316)
        at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:136)
[...]

PR #139, opened 2020-01-18
rethrow more exceptions when finalizing a ws connection

amesgen:
When closing the WS connection in the resource finalizer there might be an exception. Maybe there is an additional, helpful exception stuck in the queue (received in `onError` in the JDK WS listener), so we rethrow it. So this change improves diagnostics in some rare cases.

Initially discussed in #116

### DIFF
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWSClient.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWSClient.scala
@@ -1,14 +1,18 @@
 package org.http4s.client.jdkhttpclient
 
+import java.io.IOException
 import java.net.URI
 import java.net.http.{HttpClient, WebSocket => JWebSocket}
 import java.nio.ByteBuffer
 import java.util.concurrent.{CompletableFuture, CompletionStage}
 
 import cats._
+import cats.data.NonEmptyList
 import cats.effect._
 import cats.effect.concurrent.Semaphore
+import cats.effect.util.CompositeException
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/client/jdkhttpclient/JdkWSClient.scala
```diff
@@ -83,13 +87,23 @@ object JdkWSClient {
...
               for {
                 isOutputOpen <- F.delay(!webSocket.isOutputClosed)
                 closeOutput = fromCompletableFuture(
                   F.delay(webSocket.sendClose(JWebSocket.NORMAL_CLOSURE, ""))
                 )
-                _ <- closeOutput.whenA(isOutputOpen)
+                _ <- closeOutput.whenA(isOutputOpen).onError {
+                  case e: IOException =>
+                    for {
+                      chunk <- queue.tryDequeueChunk1(10)
```
amesgen: This `10` is arbitrary, `1` should be enough, but I think it doesn't hurt.

amesgen:
travis failure is unrelated..............

rossabaker:
Goodness, those 403s are irritating.  I wonder whether GitHub Actions suffers as much?

amesgen:
I am actually motivated to try to migrate this repo to GH Actions CI (prbbly tomorrow).

rossabaker:
In addition to being an excellent module in its own right, this repo has been a nice proving ground for other improvements.  Would love to see that tried here.

PR #141, opened 2020-01-19
add github actions CI

amesgen:
Motivated by #139

Does not include snapshot publishing for now, as I cannot add secrets to this repo.


amesgen:
Seems to work: https://github.com/amesgen/http4s-jdk-http-client/commit/e006226c2b2b98d71fad014a752e2e2b211caea6/checks

amesgen:
Thanks, but I forgot that I don't actually know the secrets (Travis does not display them) ^^

Can you add them at https://github.com/http4s/http4s-jdk-http-client/settings/secrets (same key-value format)?

I am not entirely sure if snapshot publishing works (I can't really test it), but the tests should work fine.

rossabaker:
I added the secrets.  If any maintainers have 1Password, I can share all but the Sonatype.

amesgen:
Hmm, snapshot publishing [fails](https://github.com/http4s/http4s-jdk-http-client/runs/402084912#step:6:16), can you double-check the values?

I don't have 1Password, but this procedure should be rare enough (hopefully...)

rossabaker:
I don't remember telling it anything about a GPG public key.  Let me figure out where that goes.

PR #143, opened 2020-01-22
Release v0.2.0-RC1

rossabaker:


amesgen:
Hmm, the new CI config now runs the checks twice if one creates a PR from a branch in this repo. I think that is unavoidable right now, as IMO running CI in other branches is a good thing.

rossabaker:
Maybe I should get back out of the habit of pushing directly to the main repo.  I'm used to the private repos at work that don't have forks, and am not toggling between OSS mode and work mode very well.

PR #144, opened 2020-01-22
Upgrade to http4s-0.21.0-RC1

rossabaker:
We missed the most important upgrade of all. I'll release this as 0.2.0-RC2.


rossabaker:
Yep, I noticed the miss in #143 when I looked at the docs and wondered why they didn't update. :smile:

PR #149, opened 2020-01-28
Match the SBT Scala Versions to GitHub Actions

rossabaker:


rossabaker:
I agree.  Does sbt-travisci still work, or is there a better way now?

PR #153, opened 2020-02-01
Use sonatypeBundleRelease as a release step

rossabaker:
I've had to do this manually the last couple times.


rossabaker:
Yeah, this is a lot faster, and we're already publishing to the local bundle, so we need to release the local bundle.  The mismatch has led me to believe I released twice without getting the job done.

PR #155, opened 2020-02-01
Unescape the quotations in ssh-agent eval

rossabaker:
Shell scripts in YAML, what could go wrong?


rossabaker:
We've got the same goal here. :smile:

PR #157, opened 2020-02-02
fix site publishing

amesgen:
Site publishing did not work previously (see [gh-pages log](https://github.com/http4s/http4s-jdk-http-client/commits/gh-pages), [error message](https://github.com/http4s/http4s-jdk-http-client/runs/421426336?check_suite_focus=true#step:6:119)).

I am not sure why this fails (the job does not fail as sbt-ghpages [swallows the error](https://github.com/sbt/sbt-ghpages/blob/master/src/main/scala/com/typesafe/sbt/sbtghpages/GhpagesPlugin.scala#L95-L98)), but this workaround does work, and I think it is in fact "better" that local publishing uses the local github user information and not "GitHub Actions CI". The commit message for local publishing will be "updated site" ([src](https://github.com/sbt/sbt-ghpages/blob/master/src/main/scala/com/typesafe/sbt/sbtghpages/GhpagesPlugin.scala#L85)).


amesgen:
By the way: 0.2.0-RC{2,3} are missing in gh-pages, sould site publishing be part of the release process?

PR #158, opened 2020-02-02
simplify gpg workaround

amesgen:
see the test-publish branch


amesgen:
At least we archived the workaround in the git history ;)

PR #160, opened 2020-02-03
add site publishing as a release step

### LINKED ISSUE #146: sbt-release is not actually releasing
I thought I released last night, but I had do do an additional `sonatypeBundleRelease`.  Would also be nice if it published the site, which is also a separate step.

amesgen:
Trivial change to close #146 (in combination with #153). I can't test this, but it seems fine.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,7 @@ lazy val releaseSettings = {
       tagRelease,
       // For non cross-build projects, use releaseStepCommand("publishSigned")
       releaseStepCommandAndRemaining("+publishSigned"),
+      releaseStepCommand("docs/ghpagesPushSite"),
       setNextVersion,
       commitNextVersion,
       releaseStepCommand("sonatypeBundleRelease"),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -95,6 +95,7 @@ lazy val releaseSettings = {
       tagRelease,
       // For non cross-build projects, use releaseStepCommand("publishSigned")
       releaseStepCommandAndRemaining("+publishSigned"),
+      releaseStepCommand("docs/ghpagesPushSite"),
       setNextVersion,
       commitNextVersion,
       releaseStepCommand("sonatypeBundleRelease"),
```
amesgen: Is this step necessary? Snapshot publishing *should* be done via CI, isn't it?

rossabaker: This is how we publish releases to Sonatype, not snapshots, so yes.

I've been running `sbt release` locally.  Ideally, we want to do that step from CI as well.  That's under discussion on #28, which I need to go reread myself. :smile:

PR #170, opened 2020-02-09
change caching strategy & use actual build number

amesgen:
- We now use `github.run_id` instead of `github.action`
 - We use a weekly cache that uses the cache of the week before. This way, we always have a reasonably fresh cache which neither does grow without bounds nor does reset if any sbt or project/*.scala file is changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,16 +16,22 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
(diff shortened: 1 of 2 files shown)


amesgen:
Yeah, unbounded cache growth should not be a huge problem (github allows 5GB), except that restoring the cache might take some time (right now we have 35s for 219MB [src](https://github.com/http4s/http4s-jdk-http-client/runs/433636424?check_suite_focus=true#step:4:1))

PR #172, opened 2020-02-09
use Dhall for CI configuration

amesgen:
Testing ground for potential usage of Dhall for CI configuration.

One change that might also be interesting for http4s is that we use the *experimental* client mode for SBT, so that creating a single step for every SBT task does not result in long SBT waiting times. [see here](https://github.com/amesgen/http4s-jdk-http-client/runs/434991508)


rossabaker:
Neat.

Things I like:
- It's much simpler than what we have.
- It might also be used to generate a mergify config,  since we need to enumerate all the jobs in the CI action for mergify.

Things I don't like:
- We have to check in generated source. We should consider a workflow to run the dhall and fail if it detects a difference.
- I dislike having non-JVM dependencies.  This might give me the kick in the ass to revive scala-dhall.
- Other contributors have to buy into Dhall, or the pool of build maintainers shrinks.  And I don't think the status quo is _that bad_ yet.  But I can see it getting that bad as there are more repos and builds grow more tentacles.
 
I'd like to see what others think, because what we do here, I'd like to do in all our repos.

amesgen:
A `.mergify.yml` (template from http4s) is now being generated.

As you suggested, I created a CI check to ensure that the generated YAML files are kept up to date (see [here](https://github.com/amesgen/http4s-jdk-http-client/runs/442269521) for a failure). I am in favor of also running dhall-format, but I don't know if we want to format to "ASCII" or "Unicode".

build.sbt
```diff
@@ -41,12 +41,15 @@ val javaWebsocketV = "1.4.0"
 val kindProjectorV = "0.10.3"
 val betterMonadicForV = "0.3.1"
 
+lazy val scalaVersions = scala.io.Source.fromFile("scalaVersions.dhall")
+  .mkString.replaceAll("""\[|\]|"| """, "").trim.split(",")
```
LeifW: `scalaVersions.dhall` is also valid JSON - not sure if that's something we'd want to rely on, though.

amesgen: Yeah, am also not sure about this.

Obviously, it would be best to use a Dhall implementation for parsing, but [dhall-eta](https://github.com/eta-lang/dhall-eta), the only JVM implementation, uses [eta](https://github.com/typelead/eta), which is unmaintained.

Another option would be to move `scalaVersions.dhall` to the `dhall` directory and generate a `scalaVersion.json` from it—this way, we could use all Dhall features in it, and do not rely on it accidentally being JSON. Actually, I think I slightly prefer this option—what do you think?

rossabaker: I've toyed with reviving the scala-dhall project, but there's this pesky open source HTTP library that's eating up all of my open source time at the moment.

rossabaker:
In addition to the mergify maintenance, another real-world argument for this is that our scalafmt silently stopped running in the core repo when we upgraded our actions and changed the Java version naming scheme.

build.sbt
amesgen: I would be interested in contributing to a scala-dhall, but a fully compliant implementation is indeed a big project, and I don't think it is worth to also reimplement `dhall format` and `dhall-to-yaml` in Scala, even though it is indeed sad to have non-JVM dependencies.

rossabaker:
Travis Brown is working on a JVM dhall implementation, which could reduce the external tooling burden and mitigate a lot of the downside here.

amesgen:
That would be awesome! Just for clarity, I want to mention that even right now, no Dhall installation is needed for starting sbt or editing scala files.

hamnis:
I really hate yaml, so anything that reduces manual maintenance of that I like. 
I am however, not a fan of unicode in source. Having the lambdas and unicode arrows makes it less accessible IMHO. alternative syntax is `\(foo: Bar)` and normal arrows `->`

amesgen:
@hamnis Yes, we can switch to ASCII formatting, no problem! I mentioned this [here](https://github.com/http4s/http4s-jdk-http-client/pull/172#issuecomment-585446378), and as I am personally indifferent on ASCII vs Unicode, I chose the default. Note that writing Dhall does *not* require using λ and →, the formatter "upgrades" them if desired.

This issue is related: dhall-lang/dhall-haskell#1610

I will change the formatting to ASCII for now unless someone has a strong opinion for Unicode formatting.

LeifW:
I think the cons include having to install a non-JVM dependency to regenerate these files (though the generated yaml will be checked-in, so installing dhall shouldn't be needed on CI), and it being a new language for potential contributors to learn.
I imagine the set of people editing the CI setup for this will be small, so it doesn't seem like a big deal - still, it's another barrier. But maybe dhall will become "the standard" for things like this?
This sort of thing (generating YAML from the data structures of a typed language) could also be done with e.g. Scala.
I translated the `CI` type here: https://github.com/vmchale/github-actions-dhall/blob/master/haskell-ci.dhall#L43
to Scala; so this example instance of that type: `CI(name = "bob", jobs = Job(Build("ubuntu", List(Name("Build", "make install"), Uses("bob", None)))))`
https://github.com/LeifW/ScalaGithubActions/blob/master/src/main/scala/CI.scala#L70
generates the following YAML:
```yaml
name: bob
'on':
- push
jobs:
  build:
    runs-on: ubuntu
    steps:
    - Name:
        name: Build
        run: make install
    - Uses:
        uses: bob
```

amesgen:
Sorry if I merged this too quickly!

As @rossabaker mentioned, we will probably have a Dhall implementation on the JVM in the near future.

My main reason against using Scala for generating the YAML files is that one would have to setup a new repo with CI and publishing in order to share this code between different projects — with Dhall, we can just use the HTTP import feature. I did not have the time to look into e.g. https://github.com/djspiewak/sbt-github-actions , I think it is promising.

rossabaker:
No worries about merging.  We've been staring at it for a while, and I'm excited to see whether it works well or not in practice.

PR #186, opened 2020-03-02
add mergify for scala-steward

amesgen:
"Backported" from #172


rossabaker:
Overlooked this was still out there.  I'll poke gitter about #172 again, as well.

PR #187, opened 2020-03-03
mention restricted headers

amesgen:
Motivated by [this](https://gitter.im/http4s/http4s?at=5e5e5f41b873303e27785323).

Editing this markdown file is a pain due to articifial line breaks, do we want to do something about it (-> long lines)?


rossabaker:
I've started doing sentence-per-line in my Markdown.  Word-wrap runs into the problem you mention, and long lines aren't great for diffs.  Sentence-per-line is pleasing from a git perspective, but not so pleasing visually.  I don't have a great answer here.

PR #191, opened 2020-03-19
add shifting via ContextShift

amesgen:
Motivated by http4s/http4s#3236.

The package object can be removed once http4s/http4s#3255 is merged.

The MIMA failure is expected.


djspiewak:
The shifting is important, obviously, but you can also get the same effect by using the same trick as `invokeCallback` and routing through `F.start`. That's obviously super-hacky, but it does work.

amesgen:
Ah, nice to know. I think it is fine to break bincompat in http4s-jdk-http-client for the proper fix, but that technique seems to very useful to fix the issue in a patch release for http4s-async-http-client.

amesgen:
How do I get CI on green? I can set the version to `0.3.0-SNAPSHOT` in version.sbt, should I do this in this PR?

rossabaker:
Yes.  I thought we were on sbt-ci-release, which doesn't have a version.sbt, but apparently we haven't upgraded this project yet.  I don't know how we solve it when sbt-ci-release's git versioning takes over, but that's a problem for the future.

amesgen:
Yes, that is the first downside of sbt-ci-release that I became aware of.

Do we want to wait with this PR until http4s 0.21.2 has been released?

rossabaker:
Yes, that release should be imminent, so no real reason to introduce the package object for just a few days.

PR #205, opened 2020-04-02
update changelog for 0.3.0

amesgen:


amesgen:
We probably want to include #207.

rossabaker:
Okay, I think we're ready t go, right?

rossabaker:
Synchronizing to Maven Central now.

amesgen:
Am I forgetting something or are the usual "Setting version to ..." commits missing (might be annoying for the next version release)?

Related: Do we want to change this repo to sbt-ci-release? There is [this issue](https://github.com/http4s/http4s-jdk-http-client/pull/191#issuecomment-601974675), but we could work around it with manual mima exclusions, so I am in favor of sbt-ci-release, as manual publishing is a pain.

rossabaker:
I tried running `sbt release`, but it crashed on the doc publication, and I neglected the subsequent steps.  I'll clean that up.

And :+1: to converting.  I want to get out of the manual publication business.

Issue #200, opened 2020-04-01
Deadlock eventually occurs in `JdkHttpClient` for certain requests


Billzabob:
I am currently sharing the Java 8 HTTP client to create both the WebSocket client and the HTTP client. I make a single request with the HTTP client followed by connecting to a WebSocket and sending messages periodically. Eventually (always around 400 seconds later), everything hangs and CPU usage is locked at 100%.

This doesn't occur if:
- I create a separate client and WebSocket client using `.simple[IO]`
- I don't make the initial request with the client
- I request other URLs with the client

Since it doesn't happen with every URL, I'm assuming it has something to do with the endpoint I'm hitting. It gives the following response:
```
Response(
  status=200,
  headers=Headers(
    cf-cache-status: DYNAMIC,
    cf-ray: 57d4a1eb4eaaa062-SLC,
    connection: keep-alive,
    content-type: application/json,
    date: Wed, 01 Apr 2020 19:14:20 GMT,
    expect-ct: max-age=604800,
    report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct",
    server: cloudflare,
    set-cookie: <REDACTED>,
    set-cookie: <REDACTED>,
    strict-transport-security: max-age=31536000; includeSubDomains,
    transfer-encoding: chunked,
    via: 1.1 google,
    x-envoy-upstream-service-time: 26,
    x-ratelimit-bucket: 41f9cd5d28af77da04563bcb1d67fdfd,
    x-ratelimit-limit: 2,
    x-ratelimit-remaining: 1,
    x-ratelimit-reset: 1585768466,
    x-ratelimit-reset-after: 5
  )
)

{
  "url": "wss://gateway.discord.gg",
  "shards": 1,
  "session_start_limit": {
    "total": 1000,
    "remaining": 992,
    "reset_after": 23572680
  }
}
```

I tried to put together a minimized example of what I'm seeing here:
```scala
import cats.effect._
import cats.implicits._
import fs2.Stream
import java.net.http.HttpClient
import org.http4s.client.jdkhttpclient._
import org.http4s.client.jdkhttpclient.WSFrame._
import org.http4s._
import org.http4s.headers._
import org.http4s.implicits._
import scala.concurrent.duration._

object Main extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
[...]

PR #207, opened 2020-04-03
fix connection leak if body is not consumed

amesgen:
A bit surprising that noone ran into this before.

Thanks @Billzabob for discovering this [here](https://github.com/http4s/http4s-jdk-http-client/issues/200#issuecomment-608196475)!

### DIFF
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
@@ -1,10 +1,5 @@
 package org.http4s.client.jdkhttpclient
 
-import cats.ApplicativeError
-import cats.effect._
-import cats.implicits._
-import fs2.interop.reactivestreams._
-import fs2.{Chunk, Stream}
 import java.net.URI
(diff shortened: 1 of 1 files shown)


amesgen:
This issue does not seem to fix the `too many concurrent streams` issue I linked to, but I think this PR makes sense either way.

amesgen:
This does *not* fix the `too many concurrent streams` in #200, but I think it is useful either way.

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -1,17 +1,19 @@
...
 import java.nio.ByteBuffer
 import java.util
 import java.util.concurrent.Flow
+
+import cats.ApplicativeError
+import cats.effect._
+import cats.implicits._
+import fs2.concurrent.SignallingRef
+import fs2.interop.reactivestreams._
+import fs2.{Chunk, Stream}
```
rossabaker, line 16: Not a big deal, but I generally try to keep these alphabetized, since all editors can do it easily.

PR #213, opened 2020-04-09
use sbt-explicit-deps in CI

amesgen:


build.sbt
```diff
@@ -38,6 +39,26 @@ val betterMonadicForV = "0.3.1"
...
+  "org.http4s" %% "http4s-client" % http4sV,
+  "org.reactivestreams" % "reactive-streams" % reactiveStreamsV,
+  "io.chrisdavenport" %% "vault" % vaultV,
+  "org.http4s" %% "http4s-testing" % http4sV % Test,
+  "org.specs2" %% "specs2-core" % specs2V % Test,
+  "org.specs2" %% "specs2-scalacheck" % specs2V % Test,
+  "com.codecommit" %% "cats-effect-testing-specs2" % catsEffectTestingV % Test,
+  "org.http4s" %% "http4s-dsl" % http4sV % Test,
+  "org.http4s" %% "http4s-blaze-server" % http4sV % Test,
+  "org.java-websocket" % "Java-WebSocket" % javaWebsocketV % Test
```
rossabaker, line 59: I might alphabetize these, just to provide some order.

amesgen:
CI conf has also been touched in #201, where I found it very useful (espc. mergify config). In this PR, the only Dhall "benefit" is the DRYness w.r.t. writing `sbt ++$SCALA_VERSION`. Personally, I like it so far.

Like I said [here](https://github.com/http4s/http4s-jdk-http-client/pull/172#issuecomment-607430810), [sbt-github-actions](https://github.com/djspiewak/sbt-github-actions) might be interesting to explore.
With it, no Dhall knowledge is required for contributing, but it is not clear to me how easy it is to generate e.g. the mergify conf, and sharing config snippets gets a bit more difficult.

rossabaker:
Ah, right, I forgot about that other one.  My number of projects has increased as my brainpower has decreased!

I like what you've done better than sbt-github-actions.  I'm in favor of continuing to run with it for a bit and see how it wears over time.

build.sbt
```diff
@@ -38,6 +39,29 @@ val betterMonadicForV = "0.3.1"
...
+) ++ Seq(
+  "com.codecommit"      %% "cats-effect-testing-specs2" % catsEffectTestingV,
+  "org.http4s"          %% "http4s-blaze-server"        % http4sV,
+  "org.http4s"          %% "http4s-dsl"                 % http4sV,
+  "org.http4s"          %% "http4s-testing"             % http4sV,
+  "org.java-websocket"  %  "Java-WebSocket"             % javaWebsocketV,
+  "org.specs2"          %% "specs2-core"                % specs2V,
+  "org.specs2"          %% "specs2-scalacheck"          % specs2V,
+).map(_ % Test)
+// format: on
```
amesgen, line 63: Would be nice to tell scalafmt to align this section, but I think there are no local options.

amesgen: Hmm, I can't seem to get this to work. I now have 

```scala
// scalafmt: { align = most, trailingCommas = always }
val coreDeps = ???
// scalafmt: { align = none, trailingCommas = never }
```
and nothing changes...

But I think it is fine to merge this PR even without this extremely important formatting functionality :)

rossabaker: We don't even have scalafmt on in the main repo, because it would be a pain across maintenance branches.  I'd like to resolve that someday, but again, we've lived without. :smile:

amesgen: Yeah, it is also not checked by scalafmtCheckAll, but it is certainly not very important..

I created scalameta/scalafmt#1861 for the formatting issue, but this PR should be ready now.

PR #216, opened 2020-04-14
dhall-haskell → dhallj

amesgen:
We should at least wait for the next dhallj release.

Two minor downsides:

 - No formatting check right now, but this might change.
 - The generated YAML has ugly multi-line strings, and I think that is a limitation of SnakeYAML (see [here](https://bitbucket.org/asomov/snakeyaml/issues/304/unable-to-produce-folded-block-for-multi)).

Just run `convertDhall` for updating the YAML files.


LeifW:
Not sure what you mean about multi-line strings - the output here looks like straight JSON?
I get that these files aren't intended to be edited, but it would be nice to have them a bit more readable.
When I used circe-yaml it generated nicely formatted YAML output, and it uses SnakeYAML, so hopefully the same would be possible here?
https://github.com/http4s/http4s-jdk-http-client/pull/172#issuecomment-607424549

amesgen:
@LeifW Sorry, I am messing around right now (I will rebase these commits). Now, I use YAML again, the multi-line strings in e.g. release.yml look very ugly, the rest is fine.

amesgen:
The github UI is confused, I will create another PR.

PR #217, opened 2020-04-14
dhall-haskell → dhallj

amesgen:
With [dhallj](https://github.com/travisbrown/dhallj), we no longer depend on dhall-haskell and its binaries, in order to only depend on a JVM.

~Two~ one minor downside:

 - No formatting check right now, but this might change.
 - ~The generated YAML has ugly multi-line strings, and I think that is a limitation of SnakeYAML (see [here](https://bitbucket.org/asomov/snakeyaml/issues/304/unable-to-produce-folded-block-for-multi)).~ ~This is a known issue in dhallj, will probably be fixed soon.~ Fixed in dhallj 0.2.0.

Just run `convertDhall` for updating the YAML files.


rossabaker:
Eliminating the `curl` calls in the build is a big plus, as it also means simplifying (build) developer environments. This is good when it's ready.

amesgen:
This PR should now be ready. I updated the PR description.

dhall/ci.dhall
```diff
@@ -24,7 +24,7 @@ let steps =
                 c.Run::{
                 , name = "Test docs"
                 , run = "docs/makeSite"
-                , if = Some "startsWith(matrix.scala, '2.12')"
+                , `if` = Some "startsWith(matrix.scala, '2.12')"
```
amesgen, line 27: These backticks are necessary as the previous version was only accepted due to a bug in dhall-haskell.

amesgen:
Ah, release.yml still has one issue, " is escaped to \".

amesgen:
Ok, this is in fact a dhallj issue, I will report it. I personally don't think it is worth to temporarily revert this PR, as this does not seem like a major issue.

travisbrown:
Sorry about these escaping issues—the YAML export clearly needs some attention and tests, which I should be able to get to in the next few days.

PR #237, opened 2020-06-30
only require Java 8 for the sbt project

amesgen:
Closes fthomas/scala-steward#1512


amesgen:
I am not sure if we want to add a CI job just to check that the sbt project can be loaded with java 8...

PR #251, opened 2020-07-13
update scala to 2.12.12 and 2.13.3

amesgen:
metals does not support 2.12.12, but it does support 2.13.3, so I think it is fine to merge this.


rossabaker:
Hmm, 2.12.12 escaped my notice. :+1:

PR #261, opened 2020-08-10
changelog for 0.3.1

amesgen:


rossabaker:
Thanks, @amesgen. I'll tag it now.

PR #296, opened 2020-10-28
Add The Ability To Set Connect And Request Timeouts

isomarcte:
This commit adds the ability to set connect and request timeouts. Technically, connect timeouts could already be set before this by creating a jdk `HttpClient`, setting the connect timeout directly and passing it to the `JdkHttpClient.apply` function. However, because the JDK `HttpClient` sets the request timeout _per request_ on the `HttpRequest.Builder`, there was no way to set a request timeout.

Unfortunately, adding the ability to set these timeouts to the existing `apply` method would break binary compatibility. To avoid that, this commit also adds a `JdkHttpClientBuilder`, which allows for adding more settings in the future via this interface without breaking binary compatibility. Effectively the same as is done in the core `http4s` code, e.g, `BlazeClientBuilder`, etc.

In an effort to keep the API simple, the `JdkHttpClient.apply` method was deprecated. The intent is to have a single interface to set the settings and create a `Client`.

Some code was added to `JdkHttpClient` to handle converting between Scala and Java Duration types. Some which formerly were defined inside the `JdkHttpClient.apply` method were pulled out so they could be used by `JdkHttpClient.apply` and `JdkHttpClient.fromBuilder`.

### DIFF
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
@@ -5,10 +5,13 @@ import java.net.http.HttpRequest.BodyPublishers
 import java.net.http.HttpResponse.BodyHandlers
 import java.net.http.{HttpClient, HttpRequest, HttpResponse}
 import java.nio.ByteBuffer
+import java.time.temporal.ChronoUnit
+import java.time.{Duration => JDuration}
 import java.util
 import java.util.concurrent.Flow
+import java.util.concurrent.TimeUnit
 
-import cats.ApplicativeError
+import cats._
 import cats.effect._
 import cats.implicits._
 import fs2.concurrent.SignallingRef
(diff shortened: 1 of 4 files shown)


amesgen:
Hmm, what use case do you have in mind that is not enabled by using e.g. the [`timeout`](https://typelevel.org/cats-effect/api/cats/effect/Concurrent$.html#timeout[F[_],A](fa:F[A],duration:scala.concurrent.duration.FiniteDuration)(implicitF:cats.effect.Concurrent[F],implicittimer:cats.effect.Timer[F]):F[A]) function from cats-effect?

isomarcte:
@amesgen that's a good point. It hadn't occurred to me to use `.timeout`. I _think_ that will work. Should I close?

amesgen:
Yes, I think. Sorry that your work in this PR won't be merged now, it looks quite nice! If anything does not work, feel free to reopen/comment here or on [Gitter](https://gitter.im/http4s/http4s).

PR #331, opened 2020-12-23
Terminate CompletableFuture Values In All Exit States

isomarcte:
This commit adds a new function `fromCompletableFutureShift` and replaces the usage of `fromCompletionStage` with this function.

This function is similar to `fromCompletionStage`, except that it brackets on the effect, ensuring that in the case of cancellation or error the `CompletableFuture` is terminated with `.cancel` or `.completeExceptionally`.

Failing to do this, as `fromCompletionStage` did, can lead to a resource leak and extra CPU cycles when the effect which results from `fromCompletionStage` is terminated due to some external condition, e.g. `.timeout`.

To verify this, one needs only replace `fromCompletableFutureShift` in `CompletableFutureTerminationTest` with `org.http4s.internal.fromCompletionStage` and run the test.

### DIFF
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
@@ -16,7 +16,6 @@ import fs2.interop.reactivestreams._
 import fs2.{Chunk, Stream}
 import org.http4s.client.Client
 import org.http4s.client.jdkhttpclient.compat.CollectionConverters._
-import org.http4s.internal.fromCompletionStage
 import org.http4s.util.CaseInsensitiveString
 import org.http4s.{Header, Headers, HttpVersion, Request, Response, Status}
 import org.reactivestreams.FlowAdapters
@@ -88,7 +87,9 @@ object JdkHttpClient {
       for {
         req <- Resource.liftF(convertRequest(req))
         res <- Resource.liftF(
-          fromCompletionStage(F.delay(jdkHttpClient.sendAsync(req, BodyHandlers.ofPublisher)))
+          fromCompletableFutureShift(
+            F.delay(jdkHttpClient.sendAsync(req, BodyHandlers.ofPublisher))
+          )
         )
         res <- convertResponse(res)
       } yield res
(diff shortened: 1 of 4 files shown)


isomarcte:
Hopefully the test demonstrates the issue. As additional anecdotal evidence, I've seen multiple memory leaks leading to OOM issues with applications using the http4s jdk client where some subset of the requests are terminated due to a `.timeout` invocation.

isomarcte:
@amesgen do you know when you'll have a chance to look at this? I realize it is the holidays for a lot of people, so I don't mean to rush you. It would mostly be helpful to know if you don't think you'll get to this for a week or so, so I can know whether or not I need to temporarily switch clients on some projects which are (slowly) crash looping.

amesgen (approved):
Excellent work, thanks a lot for your investigations!

Just to reiterate my understanding of this: The problem occurs if we cancel the request (in the cats-effect sense) before we had a chance to convert the `HttpResponse`. Therefore, we did neither `cancel`/`completeExceptionally` the `CompletableFuture` nor ensure that the `BodySubscriber` is finalized (which would happen if we converted the `HttpResponse`, but we had no chance to do that). This PR ensures that `cancel`/`completeExceptionally` is called in any case (which will *not* actually cancel anything), but upon trying to complete with the `HttpResponse`, the HTTP client will notice that we are no longer interested in the response and clean up any resources associated to this request.

I will now merge this and cut a new release ASAP.

isomarcte:
@amesgen thanks!

> Just to reiterate my understanding of this: The problem occurs if we cancel the request (in the cats-effect sense) before we had a chance to convert the HttpResponse. Therefore, we did neither cancel/completeExceptionally the CompletableFuture nor ensure that the BodySubscriber is finalized (which would happen if we converted the HttpResponse, but we had no chance to do that). This PR ensures that cancel/completeExceptionally is called in any case (which will not actually cancel anything), but upon trying to complete with the HttpResponse, the HTTP client will notice that we are no longer interested in the response and clean up any resources associated to this request.

That is generally my understanding. Additionally I believe if any of the three completion mechanisms are invoked before the ExecutorService ever has a chance to schedule the Future/Runnable that it would prevent execution from ever starting. That is, if you gave the JDK client a single threaded executor, and then started N requests and instantly canceled them, you'd prevent N-1 requests from ever doing any computation, but that is a pretty extreme edge case.

PR #333, opened 2020-12-31
changelog for 0.3.3

amesgen:


rossabaker (approved):
Note that http4s-v0.21.15 is about to be released, but it shouldn't matter much.

PR #335, opened 2021-01-16
Fix Resource Leak When The HTTPResponse Body Is Not Read

isomarcte:
From the JDK docs for HttpResponse#BodyHandlers.ofPublisher,

https://docs.oracle.com/en/java/javase/15/docs/api/java.net.http/java/net/http/HttpResponse.BodyHandlers.html#ofPublisher()

> When the HttpResponse object is returned, the response headers will have been completely read, but the body may not have been fully received yet. The HttpResponse.body() method returns a Publisher<List<ByteBuffer>> from which the body response bytes can be obtained as they are received. The publisher can and must be subscribed to only once.

Of particular note is the final sentence, "The publisher can and must be subscribed to only once.".

There is a much longer discussion on the semantics of this in the comments in the code, but the short version is that we must always ensure that the `Flow.Publisher` returned on the `HttpResponse` when we use `BodyHandlers.ofPublisher` is subscribed to _exactly once_, whether or not anyone reads the body. Once we subscribe, the `Flow.Subscription` must be read or canceled.

Prior to this commit, we weren't doing that in cases where the body was not read, e.g. `Client#status`. For clarification, we _were_ (and still are) interrupting the body stream via a `.interruptWhen` call, but this doesn't subscribe to the `Producer` if no one has attempted to pull from the `fs2.Stream` yet.

It is _really hard_ to write a unit test to demonstrate this behavior, but thankfully, it is _really easy_ to write some toy code which triggers it. To that end, see https://github.com/isomarcte/http4s-jdk-client-body-resource-leak , which is a repo which has a POC of this error. Running the `Main` method in there should trigger the issue in about 15 seconds. If you locally publish this commit and then re-run that with the new version, it will run forever using constant memory as intended.

For convenience, here are links to the relevant documentation or code,

* https://docs.oracle.com/en/java/javase/15/docs/api/java.net.http/java/net/http/HttpResponse.BodyHandlers.html#ofPublisher()
* https://github.com/typelevel/fs2/blob/v2.5.0/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala#L64
    * Note about how fs2 doesn't subscribe until someone starts pulling from the Stream.
* https://github.com/typelevel/fs2/blob/v2.5.0/reactive-streams/src/main/scala/fs2/interop/reactivestreams/package.scala#L55
    * Shows how to make a fs2.Stream from a Publisher and attach an effect to the subscription event. This commit uses this to ensure that subscription always happens exactly once.

### DIFF
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
@@ -89,7 +89,7 @@ object JdkHttpClient {
     // intentionally _do not_ subscribe to the publisher until the first
     // attempt to pull from the Stream.
     //
-    // https://github.com/typelevel/fs2/blob//reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala#L64
+    // https://github.com/typelevel/fs2/blob/v2.5.0/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala#L64
     //
     // In the general case, this is fine and probably even ideal. After all if
     // you are never going to pull from a Stream, why do all the booking setup
@@ -114,7 +114,7 @@ object JdkHttpClient {
     // Semaphore (it will never be released). The code for this is similar to
     // the body of the fromPublisher method in fs2.
     //
-    // https://github.com/typelevel/fs2/blob//reactive-streams/src/main/scala/fs2/interop/reactivestreams/package.scala#L55
+    // https://github.com/typelevel/fs2/blob/v2.5.0/reactive-streams/src/main/scala/fs2/interop/reactivestreams/package.scala#L55
     //
     // In the release section of bracket on the response, we also attempt to
     // acquire the single permit from the Semaphore. If we are successfully
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
@@ -92,8 +92,7 @@ object JdkHttpClient {
     // https://github.com/typelevel/fs2/blob/v2.5.0/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala#L64
     //
     // In the general case, this is fine and probably even ideal. After all if
-    // you are never going to pull from a Stream, why do all the booking setup
-    // work?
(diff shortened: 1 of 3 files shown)


amesgen:
Thank you very much! I thought I fixed this in #207, but now I see how that is not the case.

I will review this in detail tomorrow or on monday, to publish a new release ASAP.

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -58,38 +60,166 @@ object JdkHttpClient {
...
-              body = FlowAdapters
-                .toPublisher(res.body)
-                .toStream[F]
-                .interruptWhen(signal)
-                .flatMap(bs => Stream.fromIterator(bs.iterator.asScala.map(Chunk.byteBuffer)))
-                .flatMap(Stream.chunk)
-            ) -> signal.set(true)
+      Resource
+        .make(
+          (Semaphore[F](1L), responseF).tupled
```
amesgen, line 148: While using a semaphore with 1 permit works perfectly fine, I think it is somewhat nicer to use a `TryableDeferred` here, as this emphasizes that we only have one state transition (from "incomplete" to "complete".

Feel free to use/adapt [this diff](https://gist.github.com/amesgen/4cd0110da0271acdef96685edb1dc97d).

amesgen (changes requested):
Excellent job!

One thing that might be useful is to include an example of how this OOM could be reproduced if we happen to reintroduce it somehow. I agree that writing a unit test is not worth the complexity in this case, so what about adding a suitable main file which can be run manually if desired?

I used [this file](https://gist.github.com/amesgen/760f2d1ca60cf2cb4fde96a14b504ad7) (in `core/src/test/scala/org/http4s/client/jdkhttpclient`) and ran 

```
bloop run core-test --args -J-Xmx200M
```

with and without the changes in `JdkHttpClient`.

isomarcte:
Thanks @amesgen for review. I applied your diff (with one minor fix in the comments on a separate commit) and added the Main file. TryableDeferred is a much cleaner solution here.

isomarcte:
Huh....

This is odd... https://github.com/isomarcte/http4s-jdk-http-client/commits/client-body-resource-leak-fix

Github isn't detecting the new changes on my branch. I've never had that happen before. They are on my fork, but they aren't showing up here.

amesgen (approved):
Perfect, thanks for fixing the inverted comment!

amesgen:
Changelog PR open: #338

PR #340, opened 2021-01-20
use sbt-http4s-org

amesgen:
- [x] `PGP_SECRET` needs to be updated as mentioned by @rossabaker [here](https://github.com/http4s/sbt-http4s-org/pull/58#issuecomment-763258779) (~I don't have access to the PGP key~ exported the secrets and changed the format)
 - [x] update mergify


amesgen:
For some reason, aggregation does not seem to work for `mimaReportBinaryIssues` (but it does work for e.g. `undeclaredCompileDependenciesTest`).

Workaround in 1c912a534207abd7221a562f7637dd0bfe3d508d

ChristopherDavenport (approved):

rossabaker:
I think we'll need to regenerate the PGP key before the next release.  The plugin used doesn't accept a passphrase.  I'll swap that out now.

amesgen:
I already did this (see [above](https://github.com/http4s/http4s-jdk-http-client/pull/340#issue-558502336)), snapshot publishing is failing right now due to timeouts: https://github.com/http4s/http4s-jdk-http-client/runs/1746355171?check_suite_focus=true

rossabaker:
Oh, looks like someone already did.  I'll stand down.

rossabaker:
Yep, I missed the update. :+1:

amesgen:
Yeah, should have posted it as a comment, comment edits are too easy to miss. :+1:

PR #341, opened 2021-01-23
disable snapshot publishing for now

amesgen:
see djspiewak/sbt-spiewak#44

In particular, snapshots are not released as such, but rather as "normal" releases, see https://search.maven.org/artifact/org.http4s/http4s-jdk-http-client_2.12/0.3-4-1910a08/jar

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
   publish:
     name: Publish Artifacts
     needs: [build]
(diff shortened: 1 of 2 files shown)


amesgen:
websocket.org seems to be (close to) down. Probably a good idea to change the tests such that we do not rely on external services.

rossabaker (approved):
I've argued with Daniel on this before, but I'm not a fan of these snapshots going to the release repository.  :+1:

PR #343, opened 2021-01-26
reenable snapshot publishing

amesgen:
Revert #341, by inlining the "original" definition of `isSnapshot` from sbt-git: https://github.com/sbt/sbt-git/blob/c97aca2577d8ca36d04859cb1bbcc6d8cdb0bece/src/main/scala/com/typesafe/sbt/SbtGit.scala#L187-L189

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
   publish:
     name: Publish Artifacts
     needs: [build]
-    if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v'))
+    if: github.event_name != 'pull_request' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
     strategy:
       matrix:
         os: [ubuntu-latest]
--- a/build.sbt
+++ b/build.sbt
@@ -103,6 +103,8 @@ inThisBuild(
       WorkflowStep.Sbt(List("test"), name = Some("Run tests")),
       WorkflowStep.Sbt(List("doc"), name = Some("Build docs"))
     ),
+    isSnapshot :=
+      git.gitCurrentTags.value.isEmpty || git.gitUncommittedChanges.value,
     githubWorkflowPublishPostamble := Seq(
       WorkflowStep.Run(
         List("""
@@ -120,7 +122,7 @@ inThisBuild(
       )
     ),
     githubWorkflowPublishTargetBranches := Seq(
-      // RefPredicate.Equals(Ref.Branch("main")),
+      RefPredicate.Equals(Ref.Branch("main")),
       RefPredicate.StartsWith(Ref.Tag("v"))
     )
   )


rossabaker:
Is this going to publish hash snapshots to Maven Central, or old school snapshots?  I am not a huge fan of polluting Central with the hashes that some other projects do.

amesgen:
Snapshots should not go to Maven central: sbt-spiewak-sonatype uses [sbt-sonatype](https://github.com/xerial/sbt-sonatype), which says:

> If isSnapshot.value is true (e.g., if the version name contains -SNAPSHOT), publishSigned task will upload files to the Sonatype Snapshots repository without using the local bundle folder.

The version string will not contain `-SNAPSHOT`, though (but we could change that).

rossabaker:
My concern is not filling up Maven Central with hash releases like this:

https://repo1.maven.org/maven2/org/typelevel/cats-effect-kernel_2.13/

The monotonicity that was added (after I crabbed at Daniel about it) makes them less bad to depend on, but they're still irritating for other tooling in the Central ecosystem (e.g., sbt-updates finding them).  If someone really wants an intermediate version, I hope we're always in a position to oblige them with a milestone instead.

amesgen:
Yes, preventing the snapshot builds from reaching maven central was the motivation to adapt `isSnapshot` [here](https://github.com/http4s/http4s-jdk-http-client/pull/343/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R106-R107).

rossabaker (approved):
And if I had read the linked ticket, or even remembered the conversation from a few days ago, I'd have known all this without asking again. :facepalm:  Sorry I was being dense.

amesgen:
Hmm, this did not work, I somehow failed to test this locally (`isSnapshot` is *still* always false). On `main`:

```
sbt:root> show isSnapshot
[info] core / isSnapshot
[info] 	false
[info] isSnapshot
[info] 	false
sbt:root> show gitCurrentTags
[info] core / gitCurrentTags
[info] 	List()
[info] gitCurrentTags
[info] 	List()
```

vs.

```scala
isSnapshot := git.gitCurrentTags.value.isEmpty || git.gitUncommittedChanges.value
```

I am confused...

amesgen:
Now I am even more confused: I changed the definition of `isSnapshot` to

```scala
isSnapshot := {
  val currentTags = git.gitCurrentTags.value
  println(s"current tags: $currentTags, is empty: ${currentTags.isEmpty}")
  currentTags.isEmpty || git.gitUncommittedChanges.value
}
```

and now I get

```
[info] loading settings for project root from build.sbt ...
current tags: List(), is empty: false
```

amesgen:
Ah, `currentTags` is a single-element list, containing the empty string (but why is it not printed as `List("")`??). This is due to this line in sbt-spiwak: https://github.com/djspiewak/sbt-spiewak/blob/8cf1da8bd28068a034c95dab98f9357c84c52065/core/src/main/scala/sbtspiewak/SpiewakPlugin.scala#L232

PR #344, opened 2021-01-27
update to http4s 1.0

amesgen:
also cats-effect 3 and fs2 3


rossabaker:
Are we going to keep a maintenance branch?  Or two even, with the advent of http4s-0.22?  And if so, what's our numbering strategy?

amesgen:
Hmm, my gut reaction is to release this branch as 0.5 (not 1.0, as the websocket interface will change in the future in the move to http4s-core). We can use 0.4 for http4s 0.22, and otherwise only backport on demand.

core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -50,15 +49,15 @@ object JdkHttpClient {
     */
   def apply[F[_]](
       jdkHttpClient: HttpClient,
-      ignoredHeaders: Set[CaseInsensitiveString] = restrictedHeaders
-  )(implicit F: ConcurrentEffect[F], CS: ContextShift[F]): Client[F] = {
+      ignoredHeaders: Set[CIString] = restrictedHeaders
+  )(implicit F: Async[F]): Resource[F, Client[F]] = Dispatcher[F].map { dispatcher =>
```
rossabaker, line 53: I think one of our other backends (Jetty? I can't remember) called this `fromHttpClient`, since it no longer returns the type, but rather a `Resource`.  The thinking being that `apply` on a companion object should return an instance of the companion class.  But `JdkHttpClient` isn't a `Client`, so maybe that guideline isn't really relevant here.

They also call the thing that returns a resource `.resource`, but those also follow the builder pattern that this one isn't complicated enough to follow.

I'm bikeshedding.  This is probably all fine as is.

core/src/test/scala/org/http4s/client/jdkhttpclient/CompletableFutureTerminationTest.scala
```diff
@@ -18,23 +18,23 @@ package org.http4s.client.jdkhttpclient
...
-import cats.effect.concurrent._
 import java.util.concurrent.TimeUnit
 import scala.concurrent._
 import scala.concurrent.duration._
 import org.http4s._
 import org.http4s.server._
 import org.http4s.server.blaze._
 import cats.data._
-import cats.effect.testing.specs2.CatsIO
+import cats.effect.testing.specs2.CatsEffect
```
rossabaker, line 28: Probably more than you want to bite off, partiicularly in this PR, but I'm not regretting any of our moves toward munit in the main repo.

build.sbt
```diff
@@ -57,7 +60,7 @@ inThisBuild(
   Seq(
     crossScalaVersions := Seq("2.12.13", "2.13.4"),
     scalaVersion := (ThisBuild / crossScalaVersions).value.head,
-    baseVersion := "0.3",
+    baseVersion := "0.4",
```
rossabaker, line 63: I guess set this to 0.5 and make room for 0.4.

rossabaker (approved):
:tada:

core/src/test/scala/org/http4s/client/jdkhttpclient/CompletableFutureTerminationTest.scala
amesgen: Ah, thanks for reminding me. This is also a good idea as e.g. `ClientRouteTestBattery` are also munit-based.

core/src/test/scala/org/http4s/client/jdkhttpclient/JdkWSClientSpec.scala
```diff
@@ -33,162 +31,160 @@ import org.http4s.websocket.WebSocketFrame
...
 import scala.concurrent.duration._
+import munit.CatsEffectSuite
 
-class JdkWSClientSpec extends Specification with CatsEffect {
-  implicit val timer: cats.effect.Timer[IO] = IO.timer(global)
-  implicit val cs: cats.effect.ContextShift[IO] = IO.contextShift(global)
+class JdkWSClientSpec extends CatsEffectSuite {
 
-  val webSocket: WSClient[IO] = JdkWSClient.simple[IO].unsafeRunSync()
+  def webSocket = ResourceFixture(JdkWSClient.simple[IO])
```
amesgen, line 41: If I change `def` to `val` here, I get exceptions for all but the first test like this:

```
java.lang.IllegalStateException: Promise already completed.
    at scala.concurrent.Promise.complete(Promise.scala:53)
    at scala.concurrent.Promise.complete$(Promise.scala:52)
    at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:187)
    at scala.concurrent.Promise.success(Promise.scala:86)
    at scala.concurrent.Promise.success$(Promise.scala:86)
    at scala.concurrent.impl.Promise$DefaultPromise.success(Promise.scala:187)
    at munit.CatsEffectFunFixtures$ResourceFixture$.$anonfun$apply$4(CatsEffectFunFixtures.scala:50)
    at cats.effect.IOFiber.mapK(IOFiber.scala:970)
    at cats.effect.IOFiber.succeeded(IOFiber.scala:794)
    at cats.effect.IOFiber.runLoop(IOFiber.scala:257)
    at cats.effect.IOFiber.execR(IOFiber.scala:908)
    at cats.effect.IOFiber.run(IOFiber.scala:135)
    at cats.effect.unsafe.WorkerThread.runFiber(WorkerThread.scala:162)
    at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:295)
```

Is this expected?

rossabaker: That surprises me, too. That's a good upstream question.

rossabaker:
We should fork off series/0.3 and series/0.4 before this lands.

core/src/test/scala/org/http4s/client/jdkhttpclient/JdkWSClientSpec.scala
amesgen: This has been fixed 0.13.0

amesgen:
I think this should be ready to be merged, so we get snapshot builds and can iterate on it.

rossabaker (approved):

PR #345, opened 2021-01-27
filter empty tags

amesgen:
follow up to #343, see the recent comments there


amesgen:
djspiewak/sbt-spiewak#45

rossabaker:
My life is worse for having learned about this, but better for you having fixed it!

PR #347, opened 2021-01-29
append -SNAPSHOT to version

amesgen:
It turns out that (contrary to the documentation in the README), sbt-sonatype actually checks for the `-SNAPSHOT` suffix...

https://github.com/xerial/sbt-sonatype/blob/2f148bf1cbaf56fbf8a2534d9947305670dcd3f7/src/main/scala/xerial/sbt/Sonatype.scala#L110

Sorry for the CI PR churn!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,6 +108,11 @@ inThisBuild(
     ),
     isSnapshot :=
       git.gitCurrentTags.value.filter(_ != "").isEmpty || git.gitUncommittedChanges.value,
+    version := {
+      val v = version.value
+      val suffix = "-SNAPSHOT"
+      if (isSnapshot.value && !v.endsWith(suffix)) v + suffix else v
+    },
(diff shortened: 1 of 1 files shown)


rossabaker (approved):

amesgen:
Now, with 69947825ade9e955c73453231795f14100a096c3 on top, everything [seems to work](https://github.com/http4s/http4s-jdk-http-client/runs/1789749607?check_suite_focus=true#step:6:35). Maybe it is a good idea to get these changes concerning the different snapshot philosophy into sbt-http4s-org:

https://github.com/http4s/http4s-jdk-http-client/blob/69947825ade9e955c73453231795f14100a096c3/build.sbt#L109-L121

rossabaker:
I think that would be nice.  We stopped publishing regular snapshots when we adopted sbt-spiewak, but I think this gets us back in business.

PR #348, opened 2021-01-29
Shift Immediately After The F.async Call

isomarcte:
We need to shift _immediately_ after the `F.async` call.

### DIFF
--- a/core/src/main/scala/org/http4s/client/jdkhttpclient/package.scala
+++ b/core/src/main/scala/org/http4s/client/jdkhttpclient/package.scala
@@ -48,7 +48,7 @@ package object jdkhttpclient {
             case ex => cb(Left(ex))
           }
         }; ();
-      }
+      }.guarantee(CS.shift)
     }((cs, ec) =>
       (ec match {
         case ExitCase.Completed => F.unit
@@ -56,6 +56,6 @@ package object jdkhttpclient {
           F.delay(cs.completeExceptionally(e))
         case ExitCase.Canceled =>
           F.delay(cs.cancel(true))
-      }).void.guarantee(CS.shift)
+      }).void
     )
 }


isomarcte:
When I was fixing another bug, I reintroduced this one. 

I've got a instance where I'm seeing the ember server machinery getting scheduled on the jdk client thread pool.

isomarcte:
This is similar to https://github.com/http4s/http4s/issues/3236

isomarcte:
cc @ChristopherDavenport

isomarcte:
Here is an example stack trace. You can see the fs2.io.tcp calls (which are not used _at all_ by the Jdk client, ghty come from ember), getting mixed in with the jdk client stack.

```
java.nio.channels.InterruptedByTimeoutException: null
	at java.base/sun.nio.ch.UnixAsynchronousSocketChannelImpl$1.run(Unknown Source)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
	at async @ fs2.io.package$.asyncYield(io.scala:245)
	at apply @ org.http4s.client.jdkhttpclient.package$.$anonfun$fromCompletableFutureShift$3(package.scala:29)
	at guarantee$extension @ org.http4s.client.jdkhttpclient.package$.$anonfun$fromCompletableFutureShift$4(package.scala:43)
	at flatMap @ fs2.io.tcp.SocketGroup.$anonfun$apply$8(SocketGroup.scala:291)
	at flatMap @ fs2.io.tcp.SocketGroup.getBufferOf$1(SocketGroup.scala:260)
	at flatMap @ fs2.io.tcp.SocketGroup.fs2$io$tcp$SocketGroup$$read0$1(SocketGroup.scala:290)
	at *> @ org.http4s.client.jdkhttpclient.JdkHttpClient$.$anonfun$apply$11(JdkHttpClient.scala:185)
	at acquire @ fs2.Stream$.go$23(Stream.scala:1850)
	at *> @ org.http4s.client.jdkhttpclient.JdkHttpClient$.$anonfun$apply$11(JdkHttpClient.scala:185)
	at withPermit @ fs2.io.tcp.SocketGroup.fs2$io$tcp$SocketGroup$$read0$1(SocketGroup.scala:290)
	at map @ fs2.internal.CompileScope.interruptibleEval(CompileScope.scala:413)
	at flatMap @ fs2.internal.FreeC$.go$1(Algebra.scala:503)
	at flatMap @ fs2.internal.FreeC$.$anonfun$compile$7(Algebra.scala:463)
	at flatMap @ fs2.internal.FreeC$.go$1(Algebra.scala:460)
```

ChristopherDavenport (approved):

amesgen:
Hmm I was wondering whether the ["old" implementation](https://github.com/http4s/http4s-jdk-http-client/blob/fdc2e8812f/core/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala#L120_L134) of `fromCompletableFuture` would also work (as the [version in cats-effect 3 is very similar](https://github.com/typelevel/cats-effect/blob/5456b6bde031bbe69988b29b68fb3c401205fa19/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala#L24-L40)).

```scala
    fcf.flatMap { cf =>
      F.cancelable[A] { cb =>
        cf.handle[Unit]((result, err) =>
          err match {
            case null => cb(Right(result))
            case _: CancellationException => ()
            case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
            case ex => cb(Left(ex))
          }
        )
        F.delay { cf.cancel(true); () }
      }.guarantee(CS.shift)
    }
```

The tests pass for me locally.

amesgen:
But this can be discussed later, will merge now for a snapshot release.

amesgen:
[`0.3-14-85212a3-SNAPSHOT`](https://oss.sonatype.org/content/repositories/snapshots/org/http4s/http4s-jdk-http-client_2.12/0.3-14-85212a3-SNAPSHOT/) is up

amesgen:
#349

isomarcte:
> Hmm I was wondering whether the "old" implementation of fromCompletableFuture would also work (as the version in cats-effect 3 is very similar).

I'm not sure...I'd need to re-read the JRE docs for it and stare at it for a while. The semantics for this are always so mind bending.

isomarcte:
@amesgen just an update here. As you may have seen on Gitter, it appears the underlying issue @ChristopherDavenport and I were seeing was actually from the Ember server (we think).

I still think the old shifting location was wrong, though it may have been more minor than we had originally thought.

I still don't have a great explanation for that stack trace, we still saw it using Ember server and JDK client on 0.3.5, though it went away when we changed our test case to use the Jetty server with the JDK client.

isomarcte:
Related: https://github.com/http4s/http4s/pull/4301

PR #358, opened 2021-02-04
use a local echo server for testing

amesgen:
echo.websocket.org has caused more and more timeouts recently


amesgen:
Oh no, now we get different timeouts...

amesgen:
See [this branch](https://github.com/amesgen/http4s-jdk-http-client/commits/test-stability-experiment) for things I tried (which did not work so far).

amesgen:
Magically, the timeouts described in the hidden comments above no longer occur.  🎉

rossabaker (approved):

PR #396, opened 2021-03-22
0.3.x - Only use BodyPublishers.fromPublisher for nonzero body length

hamnis:


amesgen (approved):
Nice catch! For reference: This only happens when an empty body is sent *and* the content-length is explicitly set to 0, e.g.

```scala
// this works
client.expect(uri"https://example.org")
// this fails (without this PR)
client.expect[String](
  Request[IO](uri = uri"https://example.org")
    .putHeaders(Header("content-length", "0")))
```

Usually, I would add a regression test for this, but I instead opened #398 to track the efforts to make use of [`ClientRouteTestBattery`](https://github.com/http4s/http4s/blob/88fe521d1e962b8620933734a6d23fce844f7860/client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala).

PR #397, opened 2021-03-22
0.4.x - Only use BodyPublishers.fromPublisher for nonzero body length

hamnis:
Cherry picked from #396


amesgen:
I will manually merge #396 into `series/0.4` and `main`, and cut new patch releases. Thank you!

PR #421, opened 2021-04-07
fix sbt deprecations

amesgen:
https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
         run: sbt ++${{ matrix.scala }} headerCheckAll
 
       - name: Compile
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         run: sbt ++${{ matrix.scala }} headerCheckAll
 
       - name: Compile
-        run: 'sbt ++${{ matrix.scala }} test:compile'
+        run: sbt ++${{ matrix.scala }} Test/compile
```
rossabaker, line 66: This one is going to be a hard habit for me to break.

rossabaker (approved):

PR #448, opened 2021-05-28
Merge from 0.3

hamnis:


rossabaker (approved):
I like these merges.  They're a pain now, but if anything significant happens on a lower branch, we'll be glad we kept up with them now.

amesgen:
The test failures are spurious.

PR #469, opened 2021-07-31
Release v0.5.0

amesgen:


rossabaker (approved):
:tada:

rossabaker:
I'll tweet this one when it syncs, but it's in a race against my bedtime.

amesgen:
Weird, publishing fails due to an authorization error: https://github.com/http4s/http4s-jdk-http-client/runs/3207428472?check_suite_focus=true

rossabaker:
I can't remember whose creds that uses, but it worked last week, and no rights should have changed on Sonatype.  I guess we could retry it, and if it fails again, one of us can update the secrets?

amesgen:
Re-running CI did not help 😿  I will look into this tomorrow, I can't log into sonatype right now. :+1:

rossabaker:
Okay.  Then I bet it's your credentials in the project secrets, and I bet the login problem is related.  If you get stuck, we can use mine to do the release, but you'll probably want yours working for general purposes.

amesgen:
Just reran CI again without changing any secrets, and now publishing did work. 🎉

PR #571, opened 2022-02-03
Establish series/0.7

armanbilge:
I'm targeting this at series/0.5 since that's what it's based on, but it should get its own branch :)

Mostly opening for CI now.

I updated you to the latest sbt-http4s-org and migrated the site to Laika. We'll have to figure out the versioning problem later, but seeing as right now you have no landing page at all this should be an improvement 😉 

The site will need some tweaking, but you can ~preview it as long as my workspace is running~.
~https://8000-armanbilge-http4sjdkhttp-x69u1ksax0d.ws-us30.gitpod.io/#~
https://armanbilge.github.io/http4s-jdk-http-client

PR #577, opened 2022-02-16
Switch to upstream `WSClient`, adopt sbt-http4s-org

amesgen:
Cherry-picked version of https://github.com/http4s/http4s-jdk-http-client/pull/571

@armanbilge: This setup will cause the site to be published from series/0.7 and main for the time being until we set up versioned docs (see https://github.com/typelevel/sbt-typelevel/issues/100), right? This shouldn't be that bad as the APIs are almost the same (modulo being built against different http4s versions).


build.sbt
```diff
@@ -58,145 +54,33 @@ val coreDeps = Seq(
...
 )
 
-lazy val commonSettings = Seq(
-  testFrameworks += new TestFramework("munit.Framework")
-)
-
-lazy val generateNetlifyToml = taskKey[Unit]("Generate netlify.toml")
+ThisBuild / githubWorkflowJavaVersions := Seq("11", "17").map(JavaSpec.temurin(_))
+ThisBuild / tlCiReleaseBranches := Seq("main")
+ThisBuild / tlSitePublishBranch := Some("main")
```
armanbilge, line 70: You're right. Personally I would keep this set to `series/0.7` so that they don't overwrite each other. 0.7 should be fine until we get the versioned site up and running.

```suggestion
ThisBuild / tlSitePublishBranch := Some("series/0.7")
```

armanbilge:
> see https://github.com/typelevel/sbt-typelevel/issues/100

Actually, we already have enough in the latest sbt-typelevel to give this a try. I will make a PR shortly to the 0.7 branch to get the ball rolling on this.

armanbilge:
Oh darn, sbt-typelevel is set up for versioning but I actually "broke" it in the site template in sbt-http4s-org 🙃 I'll have to fix that first, oh well.

armanbilge (approved):
Looks good 👍

amesgen:
Will cut releases for 0.7.0 and 1.0-M1 tomorrow

armanbilge:
Actually, we should wait for the next http4s release b/c the websocket stuff is still package-private atm.

amesgen:
Ah, then it is a snapshot for 0.7 as well, thanks :+1:

Issue #563, opened 2022-01-24
Migration back to gh-pages


rossabaker:
The main site is down, and nobody else can see Netlify, so I migrated back to gh-pages so I'm not the single point of failure.

I've set up a CNAME, but GitHub still won't accept it:

```console
> dig jdk-http-client.http4s.org +noall +answer -t A

; <<>> DiG 9.10.6 <<>> jdk-http-client.http4s.org +noall +answer -t A
;; global options: +cmd
jdk-http-client.http4s.org. 2894 IN	CNAME	http4s.github.io.
http4s.github.io.	2894	IN	A	185.199.108.153
http4s.github.io.	2894	IN	A	185.199.109.153
http4s.github.io.	2894	IN	A	185.199.110.153
http4s.github.io.	2894	IN	A	185.199.111.153
```

The `/stable` and `/latest` redirects will no longer work.  I see a `/latest` directory in gh-pages.

If I ignore the security warning (which I expect will resolve itself), I get a 404.  We might need an index.html or some other landing page that redirects to a particular version.

amesgen:
I am working on a transition to Leika as part of https://github.com/http4s/http4s-jdk-http-client/pull/562, and Laika supports versioned documentation, so the need for redirects might go away as a side effect.

rossabaker:
Sorry to dump this on you abruptly, but everything's already broken, and I knew you were working on a refurb.  I'll keep an eye on things, and if you need DNS help or don't have enough GitHub rights, please yell.

armanbilge:
https://jdk-http-client.http4s.org/ is back :) still needs some follow-up, to link to the previous versions.

armanbilge:
Things mostly looking up at https://jdk-http-client.http4s.org :)

The one big surprise was that the 1.x site encourages you to install 0.3.4! I think this is because you are using squash merges rather than history-preserving merges.

amesgen:
Hmm, how can this influence which git tag is considered to be the latest one?

armanbilge:
It's because the history on main doesn't include any tags between `0.3.4` and `0.6.0-M?`.

For example, if we look https://github.com/http4s/http4s-jdk-http-client/commit/98082981e908cb31ff38aed3f60b73f97e1e9e3a, which is the commit tagged `v0.5.0` we see that it exists only on the series/0.5 and series/0.7 branches. It was "squashed" into main, so it lost the true history.

amesgen:
[...]

PR #582, opened 2022-03-02
Setup versioned documentation

armanbilge:
Towards https://github.com/http4s/http4s-jdk-http-client/issues/563. /cc @rossabaker

I pushed this to the gh-pages on my fork, have a look:
https://armanbilge.github.io/http4s-jdk-http-client

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,8 +78,8 @@ lazy val docsSettings =
         Versions(
           currentVersion = Version("0.7.x", "0.7"),
           olderVersions = Seq(
-            Version("0.5.x", "0.5"),
-            Version("0.4.x", "0.4")
+            Version("0.5.x", "0.5.0"),
+            Version("0.4.x", "0.4.0")
           )
         )
       )
--- a/build.sbt
+++ b/build.sbt
@@ -67,11 +67,23 @@ ThisBuild / developers := List(
 
 ThisBuild / githubWorkflowJavaVersions := Seq("11", "17").map(JavaSpec.temurin(_))
 ThisBuild / tlCiReleaseBranches := Seq("series/0.7")
-ThisBuild / tlSitePublishBranch := Some("series/0.7")
+// ThisBuild / tlSitePublishBranch := Some("series/0.7")
+ThisBuild / tlSitePublishBranch := Some("doc/versioned-0.7")
 
 lazy val docsSettings =
   Seq(
-    fork := true,
+    tlSiteHeliumConfig ~= {
+      import laika.rewrite._
+      _.site.versions(
+        Versions(
+          currentVersion = Version("0.7.x", "0.7"),
+          olderVersions = Seq(
+            Version("0.5.x", "0.5"),
+            Version("0.4.x", "0.4")
+          )
+        )
+      )
+    },
     mdocVariables ++= Map(
       "HTTP4S_VERSION" -> http4sV,
       "HTTP4S_VERSION_SHORT" -> http4sV.split("\\.").take(2).mkString("."),
--- a/build.sbt
+++ b/build.sbt
@@ -65,13 +65,15 @@ ThisBuild / developers := List(
   tlGitHubDev("rossabaker", "Ross A. Baker")
 )
 
+ThisBuild / tlJdkRelease := Some(11)
 ThisBuild / githubWorkflowJavaVersions := Seq("11", "17").map(JavaSpec.temurin(_))
 ThisBuild / tlCiReleaseBranches := Seq("series/0.7")
-// ThisBuild / tlSitePublishBranch := Some("series/0.7")
-ThisBuild / tlSitePublishBranch := Some("doc/versioned-0.7")
+ThisBuild / tlSitePublishBranch := Some("series/0.7")
 
 lazy val docsSettings =
   Seq(
+    tlSiteApiModule := Some((core / projectID).value),
+    tlSiteApiPackage := Some("org.http4s.jdkhttpclient"),
     tlSiteHeliumConfig ~= {
       import laika.rewrite._
       _.site.versions(
@@ -80,7 +82,8 @@ lazy val docsSettings =
           olderVersions = Seq(
             Version("0.5.x", "0.5.0"),
             Version("0.4.x", "0.4.0")
-          )
+          ),
+          renderUnversioned = false
         )
       )
     },
--- a/build.sbt
+++ b/build.sbt
@@ -80,6 +80,7 @@ lazy val docsSettings =
         Versions(
           currentVersion = Version("0.7.x", "0.7"),
           olderVersions = Seq(
+            Version("0.6.x", "0.6.0-M7"),
             Version("0.5.x", "0.5.0"),
             Version("0.4.x", "0.4.0")
           ),
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,5 +1,4 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Redirecting to https://bob.github.io/repo/</title>
 <meta http-equiv="refresh" content="0; URL=0.7">
 <link rel="canonical" href="0.7">
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,0 +1,5 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Redirecting to https://bob.github.io/repo/</title>
+<meta http-equiv="refresh" content="0; URL=0.7">
+<link rel="canonical" href="0.7">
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -67,11 +67,23 @@ ThisBuild / developers := List(
...
+      _.site.versions(
+        Versions(
+          currentVersion = Version("0.7.x", "0.7"),
+          olderVersions = Seq(
+            Version("0.5.x", "0.5.0"),
+            Version("0.4.x", "0.4.0")
+          )
+        )
+      )
+    },
```
armanbilge, line 86: This is the version config. Any versioned branch needs to include its `currentVersion`, but the `olderVersions` stuff only needs to be kept up-to-date on `main`.

Also, the old versions specifically point to `0.5.0` instead of `0.5` because historically this is how the old site was setup.

docs/index.html
```diff
@@ -0,0 +1,5 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Redirecting to https://bob.github.io/repo/</title>
+<meta http-equiv="refresh" content="0; URL=0.7">
+<link rel="canonical" href="0.7">
```
armanbilge, line 5: Woops, need to fix that title 😆 

Copied this from https://gist.github.com/domenic/1f286d415559b56d725bee51a62c24a7. I think it's the best we can do.

armanbilge:
@jenshalm sorry to be bothering you yet again 😅 if you wouldn't mind taking a quick look at what I've done here, I would appreciate it ❤️ 

The main thing I'm confused about is what is the difference between how Laika treats `index.md` and `README.md` (and `index.html` for that matter).

docs/index.html
armanbilge: Actually, I bet we can include this in a custom Helium extension :)

jenshalm:
> @jenshalm sorry to be bothering you yet again 😅 if you wouldn't mind taking a quick look at what I've done here, I would appreciate it ❤️
> 
> The main thing I'm confused about is what is the difference between how Laika treats `index.md` and `README.md` (and `index.html` for that matter).

The main difference is that `index.md` is nothing special. https://planet42.github.io/Laika/0.18/03-preparing-content/01-directory-structure.html#title-documents

armanbilge:
Thanks. So if I wanted to have an unversioned `/index.html` and a versioned `/<version>/index.html`, both the `.md` should be called `README.md`? How would I organize the directories for that?

jenshalm:
> Thanks. So if I wanted to have an unversioned `/index.html` and a versioned `/<version>/index.html`, both the `.md` should be called `README.md`? How would I organize the directories for that?

You can only have one title document per directory, as the AST contains versioned and unversioned documents in the same hierarchy (and `DocumentTree.titleDocument` is an `Option` not a `Seq`). So for your special scenario you would need to have one of them as `index.md` and live with the fact that Laika sees it as a normal document. Auto-generated navigation might not look as expected in such a case (and I feel like there is no obvious way to treat such a scenario for this feature), but everything else should just work, even though I've never personally tested this.

armanbilge:
> as the AST contains versioned and unversioned documents in the same hierarchy

Ah, when you put it like that I suppose it makes sense. The `/<version>/` is an artifact of the output format, but is not related to the hierarchy as modeled in Laika itself.

Ok, thank you very much, that was helpful.

build.sbt
```diff
@@ -67,11 +67,23 @@ ThisBuild / developers := List(
 
 ThisBuild / githubWorkflowJavaVersions := Seq("11", "17").map(JavaSpec.temurin(_))
 ThisBuild / tlCiReleaseBranches := Seq("series/0.7")
-ThisBuild / tlSitePublishBranch := Some("series/0.7")
+// ThisBuild / tlSitePublishBranch := Some("series/0.7")
+ThisBuild / tlSitePublishBranch := Some("doc/versioned-0.7")
```
rossabaker, line 71: This needs to be reverted before merge, right?

project/plugins.sbt
```diff
@@ -1 +1,3 @@
-addSbtPlugin("org.http4s" % "sbt-http4s-org" % "0.12.1")
+resolvers +=
```
rossabaker, line 1: So we're good to release 0.12.2 with this?

rossabaker (commented):
This is really not too complicated for all that it does.  I'm impressed.

build.sbt
armanbilge: Yup, that was just to get it to deploy on my fork.

project/plugins.sbt
armanbilge: Yes, good to go 👍

docs/index.html
armanbilge: We'll leave this here temporarily, since the jdk client is the only one that needs it so far, but I plan to integrate it into the next sbt-typelevel release.
- https://github.com/typelevel/sbt-typelevel/issues/199

build.sbt
```diff
@@ -65,13 +65,28 @@ ThisBuild / developers := List(
...
+    tlSiteHeliumConfig ~= {
+      import laika.rewrite._
+      _.site.versions(
+        Versions(
+          currentVersion = Version("0.7.x", "0.7"),
+          olderVersions = Seq(
+            Version("0.5.x", "0.5.0"),
+            Version("0.4.x", "0.4.0")
+          ),
+          renderUnversioned = false
```
armanbilge, line 86: We should set this to `true` on `main`.

build.sbt
```diff
@@ -65,13 +65,28 @@ ThisBuild / developers := List(
   tlGitHubDev("rossabaker", "Ross A. Baker")
 )
 
+ThisBuild / tlJdkRelease := Some(11)
```
armanbilge, line 68: New sbt-typelevel feature, to generate bytecode optimized for JDK 11.

build.sbt
```diff
@@ -65,13 +65,28 @@ ThisBuild / developers := List(
...
 ThisBuild / githubWorkflowJavaVersions := Seq("11", "17").map(JavaSpec.temurin(_))
 ThisBuild / tlCiReleaseBranches := Seq("series/0.7")
 ThisBuild / tlSitePublishBranch := Some("series/0.7")
 
 lazy val docsSettings =
   Seq(
-    fork := true,
+    tlSiteApiModule := Some((core / projectID).value),
+    tlSiteApiPackage := Some("org.http4s.jdkhttpclient"),
```
armanbilge, line 76: This points your API docs to javadoc.io.

build.sbt
```diff
@@ -65,13 +65,28 @@ ThisBuild / developers := List(
...
+    tlSiteHeliumConfig ~= {
+      import laika.rewrite._
+      _.site.versions(
+        Versions(
+          currentVersion = Version("0.7.x", "0.7"),
+          olderVersions = Seq(
+            Version("0.5.x", "0.5.0"),
+            Version("0.4.x", "0.4.0")
+          ),
```
armanbilge, line 85: Is it weird to exclude `0.6.x`? It never got a stable release in the end 🙃

armanbilge (commented):
@amesgen This PR should be good for review. It also does the sbt-http4s-org 0.12.2 update.

rossabaker (approved):

build.sbt
amesgen: Let's just add it, I don't think there is a downside.

armanbilge: 👍 I guess we can point it at the latest milestone.

armanbilge:
```suggestion
          olderVersions = Seq(
            Version("0.6.x", "0.6.0-M7"),
            Version("0.5.x", "0.5.0"),
            Version("0.4.x", "0.4.0")
          ),
```

armanbilge: FTR, it actually doesn't matter that we make this change here. All the versioning will be controlled from the main branch anyway.

amesgen (approved):
Looks great, just one question for clarification: This versioning setup requires us to manually update versions in `tlSiteHeliumConfig` and `docs/index.html` (for redirection), right? I guess some of could be automated with git tag processing, but I don't think it is a priority.

armanbilge:
> This versioning setup requires us to manually update versions in `tlSiteHeliumConfig` and `docs/index.html`

Yes, but hopefully not very often. For example you would only need to do it if you start a series 0.8. I'm going to think about how to simplify this in https://github.com/typelevel/sbt-typelevel/issues/199.

However, If you did publish an 0.5.1, you are right we would need to change it. This is because the old website was organized by patch releases, instead of minor series.

amesgen:
Great, that seems rare enough that automation is not necessary, but it seems doable if we really want it. Merging now and creating a PR to `main`.

PR #585, opened 2022-03-03
Setup versioned documentation on `main`

amesgen:


build.sbt
```diff
@@ -65,13 +65,30 @@ ThisBuild / developers := List(
...
+        Versions(
+          currentVersion = Version("1.x", "1.x"),
+          olderVersions = Seq(
+            Version("0.7.x", "0.7"),
+            Version("0.6.x", "0.6.0-M7"),
+            Version("0.5.x", "0.5.0"),
+            Version("0.4.x", "0.4.0")
+          ),
+          renderUnversioned = true
+        )
```
amesgen, line 89: @armanbilge Does this seem sensible?

armanbilge:
We can also set this one to `main` now (or just remove it, `main` is the default).
https://github.com/http4s/http4s-jdk-http-client/blob/5e2871565e2288711302fecb46e2fc4cd817de58/build.sbt#L71

build.sbt
armanbilge: Yep, that seems good 👍 I guess the only alternative to `1.x` is just `1` 😆

amesgen:
Ah yes, I forgot your comment here: https://github.com/http4s/http4s-jdk-http-client/pull/577#discussion_r808540225

.github/workflows/ci.yml
```diff
@@ -288,7 +288,7 @@ jobs:
         run: sbt '++${{ matrix.scala }}' docs/tlSite
 
       - name: Publish site
-        if: github.event_name != 'pull_request' && github.ref == 'refs/heads/series/0.7'
+        if: github.event_name != 'pull_request' && github.ref == 'refs/heads/main'
```
armanbilge, line 291: Ha, sorry :P

armanbilge (approved):
I think it's good, I guess well have to find out if it works :)

The 0.7 site versioned site is up, although sadly I see that the logo is broken for some reason 😕 I will investigate
https://jdk-http-client.http4s.org/0.7/

Note that the installation section will keep referring to 0.5.0 until we get a proper release of 0.7/1.0 out.

amesgen:
> Note that the installation section will keep referring to 0.5.0 until we get a proper release of 0.7/1.0 out.

Ah, this could be quite confusing (but can be fixed later on), I guess the easiest way to fix this is to introduce a custom mdoc variable and use it instead of @VERSION@?

armanbilge:
You can replace it with `@SNAPSHOT_VERSION@`.

amesgen:
My expectation would be that the version in the installation section points to the latest version of that series; so right now, it should point to the latest respective snapshot on both 0.7 and 1.x, and when we e.g. publish 0.7.0, it should point to that on 0.7, but still to the latest 1.x snapshot on 1.x. Is that too opinionated or can we just trust people who use prereleases to adapt this?

Either way, I am merging this PR, we can always change this later on :+1:

armanbilge:
Thanks, that's great feedback :) I had the same thoughts when working on the series/0.7 PR and so I made so changes in https://github.com/typelevel/sbt-typelevel/pull/198.

> so right now, it should point to the latest respective snapshot on both 0.7 and 1.x

The really difficult thing about snapshot versions is that the plugin doesn't really know if you published that snapshot or not. However, it is safe to assume that tags are published.

amesgen:
Looks like site publishing worked at a quick glance! Only thing that stands out to me is that the version in "Installation" is 0.3.4 here: https://jdk-http-client.http4s.org/1.x/index.html, which seems very random (it is neither the first nor the last 0.3.x release 🤔 ).

Issue #589, opened 2022-03-18
Re-enable websocket mdoc


armanbilge:
Just a reminder to activate mdoc again for the websocket stuff. I had to disable it b/c mdoc doesn't work with package privates.

amesgen:
Was closed by #592

PR #592, opened 2022-03-20
Update bundle

armanbilge:
Closes https://github.com/http4s/http4s-jdk-http-client/issues/589. Supersedes https://github.com/http4s/http4s-jdk-http-client/pull/588.

Bundles a bunch of updates and fixes up the docs for a v0.7.0 release :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,10 +18,10 @@ lazy val docs = project
   .settings(libraryDependencies ++= blazeServer)
 
 val catsV = "2.7.0"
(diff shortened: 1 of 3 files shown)


amesgen (approved):
Thanks, will also merge this into `main`.

amesgen:
@armanbilge  Do you think it is worth to set up scala-steward to also update dependencies on `series/0.7` in addition/instead of on `main`?

armanbilge:
Yes, I think that's a very good idea 👍

PR #641, opened 2022-05-22
Use Cats Effect EC as `HttpClient` executor

armanbilge:
I'm not 100% sure on this change yet, but I'm pretty sure this executor is not used for blocking operations. In which case, we should definitely prefer to use the Cats Effect compute pool here. It means one less threadpool in the application and thus less contention and less page faults, which is a big win.

```java
        /**
         * Sets the executor to be used for asynchronous and dependent tasks.
         *
         * <p> If this method is not invoked prior to {@linkplain #build()
         * building}, a default executor is created for each newly built {@code
         * HttpClient}.
```

https://github.com/openjdk/jdk11/blob/37115c8ea4aff13a8148ee2b8832b20888a5d880/src/java.net.http/share/classes/java/net/http/HttpClient.java#L253

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,8 @@ lazy val core = project
     name := "http4s-jdk-http-client",
     libraryDependencies ++= coreDeps,
     mimaBinaryIssueFilters ++= Seq(
+      // package private, due to #641
       ProblemFilters.exclude[IncompatibleMethTypeProblem](
-        // package private, due to #641
         "org.http4s.jdkhttpclient.JdkHttpClient.defaultHttpClient"
       )
     )
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import com.typesafe.tools.mima.core._
+
 lazy val root = project
   .in(file("."))
   .enablePlugins(NoPublishPlugin)
@@ -7,7 +9,13 @@ lazy val core = project
   .in(file("core"))
   .settings(
     name := "http4s-jdk-http-client",
-    libraryDependencies ++= coreDeps
+    libraryDependencies ++= coreDeps,
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleMethTypeProblem](
+        // package private, due to #641
+        "org.http4s.jdkhttpclient.JdkHttpClient.defaultHttpClient"
+      )
+    )
   )
 
 lazy val docs = project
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -252,16 +252,24 @@ object JdkHttpClient {
...
+        if (Runtime.version().feature() == 11) {
+          val params = javax.net.ssl.SSLContext.getDefault().getDefaultSSLParameters()
+          params.setProtocols(params.getProtocols().filter(_ != "TLSv1.3"))
+          builder.sslParameters(params)
+        }
+
+        ec match {
+          case exec: util.concurrent.Executor => builder.executor(exec)
+          case _ => builder.executor(ec.execute)
+        }
```
armanbilge, line 269: Hopefully we will upstream this into Cats Effect.
- https://github.com/typelevel/cats-effect/pull/2919#issuecomment-1101390355

armanbilge:
So I've been poking around in the JDK sources and again, I'm _pretty_ sure this makes sense. I think it would be strange to be running block ops on this, but I'm no expert 😅 

https://github.com/openjdk/jdk11/blob/37115c8ea4aff13a8148ee2b8832b20888a5d880/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java

build.sbt
```diff
@@ -7,7 +9,13 @@ lazy val core = project
   .in(file("core"))
   .settings(
     name := "http4s-jdk-http-client",
-    libraryDependencies ++= coreDeps
+    libraryDependencies ++= coreDeps,
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleMethTypeProblem](
+        // package private, due to #641
```
armanbilge, line 15:
```suggestion
      // package private, due to #641
      ProblemFilters.exclude[IncompatibleMethTypeProblem](
```

amesgen (commented):
I agree that it seems unplausible that the `Executor` is asked to running blocking tasks. Another point in favor of this is that by default, [`Executors.newCachedThreadPool`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html#newCachedThreadPool()) is used:

> These pools will typically improve the performance of programs that execute many short-lived asynchronous tasks.

---

Do you think it would be a good idea to update the documentation at https://jdk-http-client.http4s.org/0.7/#custom-clients to also set the executor like this? Or do we want to wait for https://github.com/typelevel/cats-effect/pull/3001?

armanbilge:
Oh, good eye! Yup, this is almost definitely the right thing to do then.

Good point, I can update the docs. We can just demonstrate it with wrapping for now, since the default work-stealing threadpool in CE isn't even an `Executor` until that PR 😅

amesgen (approved):
Approved, this can be followed up on when a release with https://github.com/typelevel/cats-effect/pull/3001 lands :+1:

PR #642, opened 2022-05-22
`fromCompletableFuture` was upstreamed

armanbilge:


amesgen (approved):
Thanks for noticing, a bit sad that we have to do such bincompat dances here, but it is not a big deal.

armanbilge:
FWIW it's not strictly necessary since it's package-private, but I like Ross's approach of only using MiMa filters when absolutely necessary.

PR #644, opened 2022-05-26
Optimizations via new entity model

armanbilge:


core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -67,18 +67,25 @@ object JdkHttpClient {
...
-                  case Some(length) if length > 0L =>
-                    BodyPublishers.fromPublisher(publisher, length)
-                  case _ => BodyPublishers.noBody
-                }
+            req.method.name,
+            req.entity match {
+              case Entity.Empty => BodyPublishers.noBody()
+              case Entity.Strict(chunk) => // TODO
+                val bv = chunk.toByteVector
+                BodyPublishers.ofInputStream(() => bv.toInputStream)
```
armanbilge, line 75: I'm not sure if an `InputStream` is the best thing to use here, but anecdotal evidence so far seems to suggest materializing an FS2 `Stream` tends to be pretty expensive.

The `Chunk -> ByteVector` conversion is unfortunate. I hope that http4s core will re-think this design choice, see:
- https://github.com/http4s/http4s/issues/5812

amesgen: What about using sth more direct, i.e.
```scala
BodyPublishers.ofByteArray(chunk.toArray)
```
or
```scala
val slice = chunk.toArraySlice
BodyPublishers.ofByteArray(slice.values, slice.offset, slice.length)
```
?

armanbilge: Aha, thank you, I didn't realize we could get access to the underlying `Array` in this way! I thought we'd be forced to copy.

```
  * In particular, calling `.toArraySlice` on a chunk returns a `Chunk.ArraySlice`, which provides
  * access to the underlying backing array, along with an offset and length, referring to a slice
  * of that array.
```

amesgen: I think it is still not guaranteed to be zero-copy, but at least it has a chance: https://github.com/typelevel/fs2/blob/ff9d758ae61cb19ad4d6f5cde4d0b236efab8157/core/shared/src/main/scala/fs2/Chunk.scala#L297-L302

armanbilge: Here's `ByteVectorInputStream` for comparison:
https://github.com/scodec/scodec-bits/blob/4ea2d16fcd54f108d37df19c76ed0e40f99351ff/core/shared/src/main/scala/scodec/bits/ByteVector.scala#L2260

Ignoring the `toBytevector` bit (since depending on our needs maybe that can change) I still wonder if it's more efficient, since the underlying client can directly copy bytes to its internal buffers regardless of how the `ByteVector` is backed. In the worst case, `toArrraySlice` will create an intermediate `Array` as you point out.

amesgen: Yeah, if we had a `ByteVector` instead of a `Chunk`, going through an `InputStream` seems like the best option. But as you say, at least for now, we have a `Chunk`, and converting `Chunk` to `ByteVector` can also copy: https://github.com/typelevel/fs2/blob/ff9d758ae61cb19ad4d6f5cde4d0b236efab8157/core/shared/src/main/scala/fs2/Chunk.scala#L365-L369

So what about going with the `ArraySlice` option for now, and change it if https://github.com/http4s/http4s/issues/5812 changes the representation?

armanbilge: Sounds good to me! :)

amesgen (approved):

PR #656, opened 2022-06-04
Update mergify config for http4s-steward

armanbilge:


build.sbt
```diff
@@ -26,7 +26,10 @@ lazy val docs = project
   .settings(libraryDependencies ++= blazeServer)
 
 ThisBuild / mergifyStewardConfig := Some(
-  MergifyStewardConfig(action = MergifyAction.Merge(method = Some("squash")))
+  MergifyStewardConfig(
+    author = "http4s-steward",
```
armanbilge, line 30: @amesgen apologies, I don't think this was right. I think it should be `http4s-steward[bot]` for mergify to pick it up ...

PR #686, opened 2022-06-24
Fix mergify config

armanbilge:
Just one of those annoying things :)


amesgen (approved):
Thanks, I forgot to follow up on https://github.com/http4s/http4s-jdk-http-client/pull/656#discussion_r889630483 :+1:

PR #695, opened 2022-07-07
docs: fix rendering of WS frame hierarchy

amesgen:
Previously:


armanbilge (approved):
Woops! One of those Laika things :)

PR #770, opened 2022-11-13
Update to Cats Effect 3.4.0

armanbilge:
The changes are a bit extensive because we no longer need to create the client inside of a `Resource`.


build.sbt
```diff
@@ -34,9 +34,11 @@ ThisBuild / mergifyStewardConfig := Some(
 ThisBuild / mergifyRequiredJobs += "site"
 ThisBuild / mergifyLabelPaths += "docs" -> file("docs")
 
+ThisBuild / resolvers ++= Resolver.sonatypeOssRepos("snapshots")
+
 val catsV = "2.9.0"
-val catsEffectV = "3.3.14"
-val fs2V = "3.3.0"
+val catsEffectV = "3.4.0"
+val fs2V = "3.4-9d4ea04-SNAPSHOT"
```
armanbilge, line 41: We will need an FS2 release, which is updated to the new `Dispatcher`s in Cats Effect 3.4.0.

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -58,35 +56,51 @@ object JdkHttpClient {
     *   cannot be set by the user. By default, the set of restricted headers of the OpenJDK 11 is
     *   used.
     */
+  @deprecated("Use `make` which does not return a `Resource`", "0.7.1")
   def apply[F[_]](
       jdkHttpClient: HttpClient,
       ignoredHeaders: Set[CIString] = restrictedHeaders
-  )(implicit F: Async[F]): Resource[F, Client[F]] = Dispatcher[F].map { dispatcher =>
```
armanbilge, line 64: We no longer need a dispatcher for the client. FS2 now uses an internal "sequential" dispatcher that can be more efficient because it is scoped to the particular stream/publisher instead of shared globally.

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -249,11 +263,17 @@ object JdkHttpClient {
...
   /** A `Client` wrapping the default `HttpClient`.
     */
+  @deprecated("Use `default` which does not return a `Resource`", "0.7.1")
   def simple[F[_]](implicit F: Async[F]): Resource[F, Client[F]] =
-    Resource.eval(defaultHttpClient[F]).flatMap(apply(_))
+    Resource.eval(default)
+
+  /** A `Client` wrapping the default `HttpClient`.
+    */
+  def default[F[_]](implicit F: Async[F]): F[Client[F]] =
```
armanbilge, line 272: Was not sure what to name it so I went with `default` /shrug

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
```diff
@@ -44,120 +43,133 @@ import java.util.concurrent.CompletionStage
 object JdkWSClient {
 
   /** Create a new `WSClient` backed by a JDK 11+ http client. */
+  @deprecated("Use `make` which does not return a `Resource`", "0.7.1")
   def apply[F[_]](
       jdkHttpClient: HttpClient
-  )(implicit F: Async[F]): Resource[F, WSClient[F]] = Dispatcher[F].map { dispatcher =>
```
armanbilge, line 49: Similar change here: instead of sharing a single, global dispatcher for the entire client, we now have an efficient "sequential" dispatcher per-connection.

armanbilge:
@amesgen this is ready to go! :)

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
amesgen: An alternative to deprecating the old `Resource`-returning functions would be to open the 0.8 release series instead, such that we don't have to invent new function names. I think I would be slightly in favor of this, as the effort for downstream users will probably be very comparable.

amesgen (approved):
Thanks, looks great! My only comment is whether going through a deprecation cycle is worth here, see the respective comment.

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
armanbilge: I had the same feeling :) ok, let's do that. I guess you should create a series/0.8 branch I can retarget to?

amesgen: > I guess you should create a series/0.8 branch I can retarget to?

https://github.com/http4s/http4s-jdk-http-client/tree/series/0.8 :+1:

PR #781, opened 2022-11-20
Update site metadata for 0.8

amesgen:


build.sbt
```diff
@@ -94,8 +94,9 @@ lazy val docsSettings =
       import laika.rewrite._
       _.site.versions(
         Versions(
-          currentVersion = Version("0.7.x", "0.7"),
+          currentVersion = Version("0.8.x", "0.8"),
           olderVersions = Seq(
+            Version("0.7.x", "0,7"),
```
armanbilge, line 99:
```suggestion
            Version("0.7.x", "0.7"),
```

armanbilge:
Also need to update this:
https://github.com/http4s/http4s-jdk-http-client/blob/a037d3af8c3763dee97e23617ed54a394d1c9194/docs/index.html#L3-L4

armanbilge (approved):
LGTM thx!

amesgen:
Hmm, the `index.html` does not seem to have been updated: https://github.com/http4s/http4s-jdk-http-client/blob/gh-pages/index.html

amesgen:
Ah, or is that only done from the `main` branch?

armanbilge:
Yup, it will update when series/0.8 merges into main.

PR #782, opened 2022-11-20
Merge `series/0.8` into `main`

amesgen:


armanbilge (approved):
CI is green. Probably fine 😅 

I mostly looked at the part where the entity model changed since I assume that had some conflicts.

PR #812, opened 2023-01-18
Migrate to new FS2 `j.u.c.Flow` interop APIs

armanbilge:
Upstream PR:
- https://github.com/typelevel/fs2/pull/3102

Previously interop was done through the reactive-streams converter, but now FS2 can interop directly with the JDK9+ interfaces. So we can remove a layer of cruft 😆

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ ThisBuild / mergifyLabelPaths += "docs" -> file("docs")
 
 val catsV = "2.9.0"
 val catsEffectV = "3.4.4"
-val fs2V = "3.4.0"
+val fs2V = "3.4.0-59-7543784-SNAPSHOT"
 val scodecV = "1.1.34"
 val http4sV = "0.23.17"
 val reactiveStreamsV = "1.0.4"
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -70,10 +68,12 @@ val coreDeps = Seq(
   "org.typelevel" %% "munit-cats-effect" % munitCatsEffectV
 )).map(_ % Test)
 
+ThisBuild / resolvers ++= Resolver.sonatypeOssRepos("snapshots")
+
 val scala213 = "2.13.10"
 ThisBuild / crossScalaVersions := Seq("2.12.17", scala213, "3.2.1")
 ThisBuild / scalaVersion := scala213
-ThisBuild / tlBaseVersion := "0.8"
+ThisBuild / tlBaseVersion := "0.9"
```
armanbilge, line 76: Because we are dropping the `reactive-streams` dependencies, Ross would tell us to bump the base version. But we are not actually breaking any APIs, only changing the internal implementations.

rossabaker: Your transitive dependencies are your API in a binary-distributed ecosystem in a broken society that rejects the good hygiene of sbt-explicit-dependencies.  I don't make the rules.

armanbilge:
This should be ready, but it needs a new `series/0.9` branch to be re-targeted to. Sorry 😅

amesgen:
@armanbilge I created [`series/0.9`](https://github.com/http4s/http4s-jdk-http-client/tree/series/0.9)

amesgen:
I think I have seen that test failure at least once before, probably flaky: https://github.com/http4s/http4s-jdk-http-client/actions/runs/4126971127/jobs/7129585543 :crying_cat_face:

armanbilge:
Yeah me too 😢 I can take a look at that in a follow-up PR.

amesgen (approved):

PR #827, opened 2023-02-08
Fix flaky ws test

armanbilge:
This should do the trick I think :)


amesgen (approved):
Thanks, rearchitecturing these tests properly such that we don't have to add random delays probably doesn't have a good power-to-weight ratio :+1:

PR #853, opened 2023-04-19
Add hard abort on resource closure if any part of the stream remains open

GrafBlutwurst:
Background of this MR is that we observed leaked sockets that eventually caused our application to crash due to exhausted file handles. Granted we use websockets a bit weirdly by having them only open a short time rather than long which probably caused this problem to show up in the first place. For more background information see: https://discord.com/channels/632277896739946517/632286375311573032/1091349652579827783

Sadly I was unable to minimize an example due to time constraint but some wireshark debugging showed the following. 

This is a healthy package exchange where all sockets in both applications (server and client) are torn down and cleaned up properly:

image

This is an unhealthy one:

image

Note that in both cases on the websocket protocol layer, close frames are exchanged properly. However in some (yet unidentified, I suspect concurrency, as the only way i was able to reproduce this is by concurrently firing many requests) cirumstances the server does not respond with a TCP FIN/ACK. 

With a blaze server this also leads to CLOSE_WAIT sockets on the server, ember seems to handles this properly as of 0.29-RC3 with the EoS fix. 

With both servers the JDK client then remains in a half open state with the server side stream remaining open. This MR checks if any side remains open in the tear down of the resource and in such a case aborts the connection. It's a bit of a sledgehammer solution but i couldn't figure out why the TCP frames are missing and this is at least a bandaid. 

The problem gets much worse if proxies are involved as the half-open state of the socket also causes the tunneling connection to remain open leading to an explosion in allocated socket that are only eventually somewhat cleaned up by JDK https internal pool handling. 

This is possibly related to https://github.com/http4s/http4s/issues/4798 though I'm entirely unsure.

### DIFF
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
@@ -129,6 +129,15 @@ object JdkWSClient {
                       })
                     } yield ()
                   }
+
+              shouldAbort <- F.delay {
+                !webSocket.isOutputClosed || !webSocket.isInputClosed
+              }
+
+              _ <- if(shouldAbort) F.delay {
+                webSocket.abort()
+              } else F.unit
+
             } yield ()
           }
           .map { case (webSocket, queue, closedDef, sendSem) =>


core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
```diff
@@ -129,6 +129,15 @@ object JdkWSClient {
...
                     } yield ()
                   }
+
+              shouldAbort <- F.delay {
+                !webSocket.isOutputClosed || !webSocket.isInputClosed
+              }
+
+              _ <- if(shouldAbort) F.delay {
+                webSocket.abort()
+              } else F.unit
```
armanbilge, line 139: I wonder if we should place this inside a `.guarantee { ... }`? And also can't we put the whole thing inside of a `delay { ... }` ?

GrafBlutwurst: Turns out you were right initially. Part of the issue is not waiting for the close frame from the server, the abort like this can lead to broken pipe errors on server side. I'll need to do a bit more thinking. In our code I'm playing around with waiting on the `closedDef` for some amount of time in case the input is still open before i do the emergency-abort.

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
```diff
@@ -47,115 +48,141 @@ object JdkWSClient {
...
-          .make {
-            for {
-              wsBuilder <- F.delay {
-                val builder = jdkHttpClient.newWebSocketBuilder()
-                val (subprotocols, hs) = req.headers.headers.partitionEither {
-                  case Header.Raw(ci"Sec-WebSocket-Protocol", p) => Left(p)
-                  case h => Right(h)
+      Dispatcher
+        .sequential(
+          await = true
```
GrafBlutwurst, line 53: This might take care of the dispatcher already shut down exception in the debug logs, haven't verified this yet. I'll check in our setup, we're using the exact same code.

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
```diff
@@ -47,115 +48,141 @@ object JdkWSClient {
...
+                  F.delay(wsBuilder.buildAsync(URI.create(req.uri.renderString), wsListener))
+                )
+                sendSem <- Semaphore[F](1L)
+              } yield (webSocket, queue, closedDef, sendSem)
+            } { case (webSocket, queue, closedDef, sendSem) =>
+              def awaitClose(step: Int): F[Unit] =
+                closedDef.tryGet.flatMap {
+                  case Some(_) => F.unit
+                  case None =>
+                    if (step < 10) F.sleep(100.millis) *> awaitClose(step + 1) else F.unit
```
GrafBlutwurst, line 118: I wonder what a sensible amount of time to wait here is? should this be configurable? overloaded `apply` with a duration parameter to retain backwards compat? In practice this alleviates the need for actually using `abort` quite a bit which is nice.

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
GrafBlutwurst: might possibly lead to non termination due to https://github.com/rivero-ag/http4s-jdk-http-client/blob/fix-hard-abort-WS-connection/core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala#L84 and the non-terminating completion stage? at least we are observing `CLOSE_WAIT` sockets again on the client.

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
armanbilge: I'm a bit confused by this, is this equivalent to `closedDef.get.timeout(1.second)` ?

GrafBlutwurst: that's what I was looking for. For some reason it wasn't in my code completion :facepalm: . thanks

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
```diff
@@ -47,115 +48,134 @@ object JdkWSClient {
...
+            }
+            .map { case (webSocket, queue, closedDef, sendSem) =>
+              // sending will throw if done in parallel
+              val rawSend = (wsf: WSFrame) =>
+                F.fromCompletableFuture(F.delay(wsf match {
+                  case WSFrame.Text(text, last) => webSocket.sendText(text, last)
+                  case WSFrame.Binary(data, last) => webSocket.sendBinary(data.toByteBuffer, last)
+                  case WSFrame.Ping(data) => webSocket.sendPing(data.toByteBuffer)
+                  case WSFrame.Pong(data) => webSocket.sendPong(data.toByteBuffer)
+                  case WSFrame.Close(statusCode, reason) => webSocket.sendClose(statusCode, reason)
```
GrafBlutwurst, line 162: One thing I have seen is this line failing because the output was already closed. We don't send close frames manually so the only way this should be invoked is by `connectHighLevel` which mirrors the close frame. Maybe we should guard this with an if, checking if the output is still open? I don't like that much though because low level usage would then have implicit restrictions. 

Also as a side note, the mirroring of the closure frame in `connectHighLevel` actually may fail if the server emits a close code that is not applicable for a client to send.

core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
armanbilge: Ah yeah, you probably just needed `import cats.effect.syntax.all._` to get that :)

Sorry, another follow-up question to this: do _we_ have to do this timeout? I am wondering if the JDK WebSocket implementation itself has an internal timeout, and we are just forgetting to listen for an error event or something. see the docs:

> Unless the CompletableFuture returned from this method completes with IllegalArgumentException, or the method throws NullPointerException, the output will be closed.
> 
> If not already closed, the input remains open until a Close message [received](https://docs.oracle.com/en/java/javase/17/docs/api/java.net.http/java/net/http/WebSocket.Listener.html#onClose(java.net.http.WebSocket,int,java.lang.String)), or abort is invoked, or an [error](https://docs.oracle.com/en/java/javase/17/docs/api/java.net.http/java/net/http/WebSocket.Listener.html#onError(java.net.http.WebSocket,java.lang.Throwable)) occurs.

https://docs.oracle.com/en/java/javase/17/docs/api/java.net.http/java/net/http/WebSocket.html#sendClose(int,java.lang.String)

ybasket:
@GrafBlutwurst Hey, sorry for the ping, but the problem this PR tackles came up again in #1015 (where I added a fix that is a simplified version of this PR). Are still planning on getting this change over the finish line? And if so, can I help somehow?

hamnis:
I tried merging series/0.9 into this PR, but too many things have changed here.
Are you able to pick this up again @GrafBlutwurst ?

GrafBlutwurst:
sadly i won't be able to get back to this pr in the forseeable future

PR #874, opened 2023-05-19
Add connection and request timeout options

hnaderi:
Added the options to control connection and request timeout.
Changes are backward binary compatible.
I added a test case for these options, but it's not working as expected in the test run, However it is working as expected using external webservers (i.e. `nc -l 8080`); so I disabled the test. I'm curious to know what's the problem, so if someone gives me a pointer I'll try to fix the tests as well.

### DIFF
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
@@ -82,7 +82,7 @@ object JdkHttpClient {
             .uri(URI.create(req.uri.renderString))
             .version(version)
 
-          requestTimeout.fold(rb)(fd => rb.timeout(Duration.ofMillis(fd.toMillis)))
+          requestTimeout.foreach(fd => rb.timeout(Duration.ofMillis(fd.toMillis)))
 
           val headers = req.headers.headers.iterator
             .filterNot(h => ignoredHeaders.contains(h.name))
@@ -275,10 +275,11 @@ object JdkHttpClient {
   ): F[Client[F]] =
     defaultHttpClient[F](connectionTimeout).map(from(_, requestTimeout))
 
+  @deprecated("use defaultHttpClient() instead", since = "0.9.2")
   private[jdkhttpclient] def defaultHttpClient[F[_]](implicit F: Async[F]): F[HttpClient] =
     defaultHttpClient(None)
 
-  private def defaultHttpClient[F[_]](
+  private[jdkhttpclient] def defaultHttpClient[F[_]](
       connectionTimeout: Option[FiniteDuration] = None
   )(implicit F: Async[F]): F[HttpClient] =
     F.executor.flatMap { exec =>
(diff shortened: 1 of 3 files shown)


armanbilge (commented):
Thanks for the PR! A couple thoughts.

1. I'm less sure about exposing the connect timeout directly, since the user already has full ability to configure the client before providing it to the http4s wrapper. We can demo this in the documentation :)

2. Regarding the request timeout: how do its semantics compare to just doing `.timeout(...)` on the `IO` returned in the http4s wrapper? I need to double-check if/how we are propagating the cancellation signal in this wrapper.

    If it makes sense to keep this configuration, I wonder if we should expose it as an attribute key on the request, so that the user can configure it per request (instead of only globally). We do something similar for all the http4s-dom per-request config.
https://github.com/http4s/http4s-dom/blob/726af7c65e4bf345b66a192bf05c5a28b2fdacd1/dom/src/main/scala/org/http4s/dom/FetchOptions.scala#L216

amesgen:
Thanks for the thoughtful comment @armanbilge! Only leaving a link to #296 for prior discussions in this area.

armanbilge:
> I need to double-check if/how we are propagating the cancellation signal in this wrapper.

Ah, I forgot that JDK APIs use `CompletableFuture` which actually has a cancellation API (unlike Scala `Future` ...). So it should Just Work:tm:.

https://github.com/http4s/http4s-jdk-http-client/blob/ae98cfdf37a95cbc6d4af244f9c494eca7834f8b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala#L233-L235

I concur with the conclusion in https://github.com/http4s/http4s-jdk-http-client/pull/296 (thanks for linking!). Actually because of the new integrated timers in CE v3.5.0 using `.timeout(...)` on `IO` will probably have better performance than doing it via the JDK APIs 😁

hnaderi:
> 1. I'm less sure about exposing the connect timeout directly, since the user already has full ability to configure the client before providing it to the http4s wrapper. We can demo this in the documentation :)

Yeah, that's just for convenience, In my experience, jdk client is really troublesome with no connection timeout. e.g. It hangs forever when the client can't establish a connection to server.

> 2. Regarding the request timeout: how do its semantics compare to just doing `.timeout(...)` on the `IO` returned in the http4s wrapper? I need to double-check if/how we are propagating the cancellation signal in this wrapper.

Actually that's the main reason for this PR and I added the connection timeout afterwards, I have experienced several times annoying differences, mostly the `.timeout` on the cats effect side doesn't work at all, unless you also specify a timeout in the jdk http request as well!
One way to produce it is the test I have added or you can also test it with a netcat session with no response. I can also provide some other examples for this later today.

>     If it makes sense to keep this configuration, I wonder if we should expose it as an attribute key on the request, so that the user can configure it per request (instead of only globally). We do something similar for all the http4s-dom per-request config.

This looks a lot better!

armanbilge:
> Actually that's the main reason for this PR and I added the connection timeout afterwards, I have experienced several times annoying differences, mostly the `.timeout` on the cats effect side doesn't work at all, unless you also specify a timeout in the jdk http request as well!

Well, that's no good! 😅 We should try and fix this bug, if possible. I will look at your reproducers a bit later. Thanks!

armanbilge:
Oh also ...

> In my experience, jdk client is really troublesome with no connection timeout. e.g. It hangs forever when the client can't establish a connection to server.

Yeah that's no good, maybe we should configure this by default in the `simple` builder. We already have a default in http4s.

https://github.com/http4s/http4s/blob/b22f722d23eabd6fc3e5e95c69166b454ee15ca3/client/shared/src/main/scala/org/http4s/client/package.scala#L25

hnaderi:
@armanbilge Thank you for your fixes :heart: , I close this one as it is not relevant anymore.

PR #875, opened 2023-05-21
Fix cancelation in http and ws clients

armanbilge:
h/t @hnaderi for reporting in https://github.com/http4s/http4s-jdk-http-client/pull/874#issuecomment-1556068884.

tl;dr we were making the requests uncancelable by mistake 🙃

### DIFF
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
@@ -41,6 +41,7 @@ import java.net.http.HttpRequest.BodyPublishers
 import java.net.http.HttpResponse
 import java.net.http.HttpResponse.BodyHandlers
 import java.nio.ByteBuffer
+import java.time.Duration
 import java.util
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -255,6 +256,7 @@ object JdkHttpClient {
         }
 
         builder.executor(exec)
+        builder.connectTimeout(Duration.ofNanos(org.http4s.client.defaults.ConnectTimeout.toNanos))
```
amesgen, line 259: Let's remove this as it is already part of #876?

amesgen (approved):
Great catch!

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
armanbilge: Woops, not sure how that slipped in :)

PR #876, opened 2023-05-21
Set default `connectTimeout` in `defaultHttpClient`

armanbilge:
Addresses https://github.com/http4s/http4s-jdk-http-client/pull/874#issuecomment-1556068884, h/t @hnaderi.


amesgen (approved):
Thanks, we could also adapt the docs of `simple` to mention this (and maybe also the already existing Java 11 TLS 1.3 workaround), as it currently still states "A `Client` wrapping the **default** `HttpClient`."

PR #980, opened 2023-10-30
remove un/rewrapping of logic with Async in defaultHttpClient

gatisnolv:
I think the 'flatMap' + subsequent 'F.delay' can be replaced by 'map'. Please correct me if I'm wrong.

### DIFF
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
@@ -253,20 +253,18 @@ object JdkHttpClient {
     defaultHttpClient[F].map(apply(_))
 
   private[jdkhttpclient] def defaultHttpClient[F[_]](implicit F: Async[F]): F[HttpClient] =
-    F.executor.flatMap { exec =>
-      F.delay {
-        val builder = HttpClient.newBuilder()
-        // workaround for https://github.com/http4s/http4s-jdk-http-client/issues/200
-        if (Runtime.version().feature() == 11) {
-          val params = javax.net.ssl.SSLContext.getDefault().getDefaultSSLParameters()
-          params.setProtocols(params.getProtocols().filter(_ != "TLSv1.3"))
-          builder.sslParameters(params)
-        }
-
-        builder.executor(exec)
-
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -253,20 +253,18 @@ object JdkHttpClient {
...
 
   private[jdkhttpclient] def defaultHttpClient[F[_]](implicit F: Async[F]): F[HttpClient] =
-    F.executor.flatMap { exec =>
-      F.delay {
-        val builder = HttpClient.newBuilder()
-        // workaround for https://github.com/http4s/http4s-jdk-http-client/issues/200
-        if (Runtime.version().feature() == 11) {
-          val params = javax.net.ssl.SSLContext.getDefault().getDefaultSSLParameters()
-          params.setProtocols(params.getProtocols().filter(_ != "TLSv1.3"))
-          builder.sslParameters(params)
```
armanbilge, line 263: Creating a new `HttpClient` is a side-effect, so it must be wrapped in `delay`.

gatisnolv: I understand it's a side-effect, but wouldn't it be suspended by residing within 'map' on the already suspended call to get the executor?

armanbilge: No, `map` cannot be used to suspend side-effects. It expects a pure function. Indeed, `map` comes from [`cats.Functor`](https://www.javadoc.io/doc/org.typelevel/cats-docs_2.13/latest/cats/Functor.html) which has no notion of side-effects at all.

It may "appear" to work, but the behavior is not guaranteed. For example, because `map` is assumed to be pure, the runtime should be safe to invoke it more than once, or to cache its result for identical inputs. All of these would break the semantics of the code.

I realize this may seem pedantic, but following these rules brings great, great benefits.
https://github.com/typelevel/cats-effect/blob/7fa9c4d703a92244c5211086518bf4cd4ce81292/README.md#five-simple-rules

PR #1015, opened 2024-02-03
Add Resource-based simple constructors

### LINKED ISSUE #1011: Java 21+: Ensure `simple` client is closed
Java 21 adds methods to close a `HttpClient` and also makes it implement `AutoCloseable`: https://bugs.openjdk.org/browse/JDK-8267140

Ideally, http4s-jdk-http-client should leverage this and expose the `simple` client wrapped in a `Resource` to ensure the client is properly closed and cleaned up after use. Not sure how this can be achieved without requiring Java 21+ though, needs research.

ybasket:
As Java 21+ allows explicitly closing clients, this adds new constructors that return the clients wrapped in a `Resource`. On older Java versions, the release is a no-op, on newer it enforces the cleanup of resources.

Also ensures the websocket client closes the input channel on release so the connection gets properly closed.

Closes #1011

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -80,7 +80,7 @@ ThisBuild / developers := List(
 )
 
 ThisBuild / tlJdkRelease := Some(11)
-ThisBuild / githubWorkflowJavaVersions := Seq("11", "17").map(JavaSpec.temurin(_))
+ThisBuild / githubWorkflowJavaVersions := Seq("11", "17", "21").map(JavaSpec.temurin(_))
 ThisBuild / tlCiReleaseBranches := Seq("series/0.9")
 ThisBuild / tlSitePublishBranch := Some("series/0.9")
 
@@ -109,8 +109,8 @@ lazy val docsSettings =
       "HTTP4S_VERSION_SHORT" -> http4sV.split("\\.").take(2).mkString("."),
       "SCALA_VERSION" -> CrossVersion.binaryScalaVersion(scalaVersion.value),
       "SCALA_VERSIONS" -> formatCrossScalaVersions((core / crossScalaVersions).value.toList)
-    ),
-    unusedCompileDependenciesFilter -= moduleFilter()
+    )
+    // unusedCompileDependenciesFilter -= moduleFilter()
   )
 
 def formatCrossScalaVersions(crossScalaVersions: List[String]): String = {
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
@@ -243,9 +243,27 @@ object JdkHttpClient {
     * [[cats.effect.kernel.Async.executor executor]], sets the
     * [[org.http4s.client.defaults.ConnectTimeout default http4s connect timeout]], and disables
     * [[https://github.com/http4s/http4s-jdk-http-client/issues/200 TLS 1.3 on JDK 11]].
+    *
+    * On Java 21 and higher, prefer [[simpleResource]] as it actively closes the underlying client,
+    * releasing its resources early.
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -243,10 +243,28 @@ object JdkHttpClient {
...
+    * releasing its resources early.
     */
   def simple[F[_]](implicit F: Async[F]): F[Client[F]] =
     defaultHttpClient[F].map(apply(_))
 
+  /** Like [[simple]], but wraps the client in a [[cats.effect.Resource Resource]] to ensure it's
+    * properly shut down on JVM 21 and higher. On lower Java versions, closing the resource does
+    * nothing (garbage collection will eventually clean up the client).
+    */
+  def simpleResource[F[_]](implicit F: Async[F]): Resource[F, Client[F]] =
```
ybasket, line 257: Name is up for discussion, I don't particular like `simpleResource`, but at least it's rather clear.

amesgen: As already mentioned in https://github.com/http4s/http4s-jdk-http-client/issues/1011#issuecomment-1919676958, we could just change the signature of `simple` and cut a 0.10 release; it should be rather simple for users to migrate.

build.sbt
```diff
@@ -1,4 +1,5 @@
 import com.typesafe.tools.mima.core._
+import explicitdeps.ExplicitDepsPlugin.autoImport.moduleFilterRemoveValue
```
amesgen, line 2: Is this due to https://github.com/cb372/sbt-explicit-dependencies/issues/33?

ybasket: Yes.

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
hamnis: I think we should rename `simple` and `simpleResource` to something like `unsafeDefault` and `default`. We should encourage users to do the right thing (using the resource variant)

We can do that in a 0.10 release, or deprecate simple in this and create new new constructors.

ybasket: @hamnis I can adjust the names, no problem, just a bit unsure about the `unsafe` prefix. On everything below Java 21, it's actually not unsafe in the sense CE usually uses this, but the only thing you can have (the `Resource` variant won't close anything there either). 
So maybe a 0.10 with `simple` (returning a `Resource`, by re-using the name users get nudged towards this) and a `def forEffect[F[_]: Async]: F[Client[F]]` (or any better name) with docs explaining the caveats on Java 21+?

hamnis: do we really need the forEffect constructor? Do we lose anything with just having the `simple` | `default` | `insertyournamehere` resource constructor ?

ybasket: @hamnis It may cause some (unnecessary) trouble for users who now need to bubble up `Resource` usage in case they hadn't so far. But as `allocated` is safe on Java below 21, I changed the PR to only offer `def simple[F[_]: Async]: Resource[F, …]` and mentioned `allocated` in the docs, so people have a hint on a workaround. Hope that's a fair compromise!

ybasket: Obviously, that means MiMa is unhappy and we need to release it as `0.10`.

hamnis (approved):

PR #1132, opened 2024-11-07
Merge to main

hamnis:
Includes #1129

No major merge conflicts arose here, all were easy to resolve.

PR #1134, opened 2024-11-07
Merge to main 

hamnis:
This is now the correct merge branch.

I apologize for my mistake.

I merged  https://github.com/http4s/http4s-jdk-http-client/pull/1132 to the wrong branch, and now we have lots of wrong commits in `series/0.9`

I reverted the merge in a follow-up PR, but not sure how or if we should resolve this.

The fix can be to rebase series/0.9 to 2b71c70c75a7abfb93166de60931e2fd27c665e7 or do nothing since the change was reverted.


rossabaker (approved):
No worries on the bad branch.  I force-pushed it back, just so the history is undisturbed.  And the commit is in the lineage of this PR, so there shouldn't be any subsequent merge pain.

PR #1146, opened 2024-12-07
Handle http/2 request body without content-length

hamnis:
Http/2 is always streaming, so setting content-length is optional.

### DIFF
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
@@ -24,6 +24,7 @@ import fs2.Chunk
 import fs2.Stream
 import fs2.concurrent.SignallingRef
 import fs2.interop.flow
+import org.http4s.EmptyBody
 import org.http4s.Header
 import org.http4s.Headers
 import org.http4s.HttpVersion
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -63,10 +63,15 @@ object JdkHttpClient {
     def convertRequest(req: Request[F]): Resource[F, HttpRequest] =
       flow.toPublisher(req.body.chunks.map(_.toByteBuffer)).evalMap { publisher =>
         convertHttpVersionFromHttp4s[F](req.httpVersion).map { version =>
+          def isStreaming = version match {
+            case HttpClient.Version.HTTP_1_1 => req.isChunked
+            case HttpClient.Version.HTTP_2 => req.contentLength.isEmpty
```
danicheg, line 68: At first glance, this seems a little tricky. But given the streaming nature, I think relying on `req.body eq Stream.empty[Nothing]` is hardly a better option.

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -63,10 +63,15 @@ object JdkHttpClient {
     def convertRequest(req: Request[F]): Resource[F, HttpRequest] =
       flow.toPublisher(req.body.chunks.map(_.toByteBuffer)).evalMap { publisher =>
         convertHttpVersionFromHttp4s[F](req.httpVersion).map { version =>
+          def isStreaming = version match {
```
danicheg, line 66: To reduce confusion, I'd suggest renaming it. One might assume that `HTTP/2` mandates the streamed body or the presence of `Content-Length` to distinguish that the body is not streamed (while the streamed body alongside `Content-Length` is ok in HTTP/2). IMHO, something implying `consumeFully` might be fine.
```suggestion
          def consumeFully = version match {
```

danicheg (commented):
Not blocking progress, just sharing some food for thought.

core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -63,16 +64,24 @@ object JdkHttpClient {
...
                 BodyPublishers.fromPublisher(publisher)
               else
                 req.contentLength match {
                   case Some(length) if length > 0L =>
                     BodyPublishers.fromPublisher(publisher, length)
-                  case _ => BodyPublishers.noBody
+                  case _ =>
+                    // If we dont do this, we might block finalization
+                    publisher.subscribe(DrainingSubscriber)
+                    BodyPublishers.noBody
```
danicheg, line 84: 👍🏻

samspills (approved):

PR #1148, opened 2024-12-10
Merge series/0.10 to main

hamnis:
A bit complicated merge in JdkHttpClient.

Please review

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,5 @@
 import com.typesafe.tools.mima.core._
+import explicitdeps.ExplicitDepsPlugin.autoImport.moduleFilterRemoveValue
 
 lazy val root = project
   .in(file("."))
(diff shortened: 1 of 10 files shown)


ybasket (approved):

build.sbt
```diff
@@ -11,10 +12,6 @@ lazy val core = project
     name := "http4s-jdk-http-client",
     libraryDependencies ++= coreDeps,
     mimaBinaryIssueFilters ++= Seq(
-      // package private, due to #641
-      ProblemFilters.exclude[IncompatibleMethTypeProblem](
-        "org.http4s.jdkhttpclient.JdkHttpClient.defaultHttpClient"
-      )
```
samspills, line 17: Why don't we need this MiMa filter in main? (I trust the MiMa check more than my brain and since it's not yelling this seems fine!, I just don't understand)

samspills (approved):

build.sbt
hamnis: MiMA in main does not make sense. We have no binary compatibility guarantees here.

Since the last merge was from 0.10, this was also removed there, since we were already starting a new series.

samspills: Ahh I see, thank you!

PR #1192, opened 2025-06-06
Makes requests uncancelable when running on Java lower than 16 where cancellation is not supported

### LINKED ISSUE #1188: Client leaks connections on cancellation
Hello there!

It appears that the `JdkHttpClient` doesn't properly release connections when a request is cancelled, leading to a leak of file descriptors.

Code to reproduce (it's slightly messy, but it does the job):
```scala
//> using dep org.http4s:http4s-jdk-http-client_2.13:0.10.0
//> using dep org.http4s:http4s-dsl_2.13:0.23.30
//> using dep org.http4s:http4s-blaze-server_2.13:0.23.17

import cats.effect._
import cats.effect.std._
import cats.syntax.all._
import org.http4s._
import org.http4s.blaze.server._
import org.http4s.client._
import org.http4s.dsl.io._
import org.http4s.jdkhttpclient._
import org.http4s.syntax.literals._
import scala.concurrent.duration._

object Main extends IOApp {
  val port: Int = 8081

  val app = HttpRoutes
    .of[IO] { case GET -> Root =>
      Ok {
        IO
          .sleep(3.seconds)
          .as("this is a response")
      }
    }
    .orNotFound
  val server =
    BlazeServerBuilder[IO]
      .bindLocal(port)
      .withHttpApp(app)
      .resource

  val request = (client: Client[IO]) =>
    client.expect[String](s"http://localhost:$port/")
  val step = (msg: String) => IO.println(s"---------- $msg ----------")

  val makeClient = JdkHttpClient.simple[IO]

  val leaks = (client: Client[IO]) =>
    step("starting leaking code") >>
      request(client).timeoutTo(1.seconds, IO.println("timed out")) >>
      step("leaking code exited")

  val doesntLeak = (client: Client[IO]) =>
    step("starting non-leaking code") >>
      request(client) >>
      step("non-leaking code exited")

  val init =
    IO.delay(java.lang.ProcessHandle.current().pid())
      .flatMap(IO.println)
      .toResource >> (server, makeClient).tupled

  def run(args: List[String]) =
    init.use { case (_, client) =>
      for {
        _ <- args.take(1) match {
          case List("leak")    => leaks(client)
          case List("no-leak") => doesntLeak(client)
          case Nil             => doesntLeak(client)
          case _ => IO.raiseError(new RuntimeException("bad arguments"))
        }
        _ <- step(
          "keeping alive for 15 seconds (connections will close after 10 s)"
        )
        _ <- IO.sleep(15.seconds)
        _ <- step("closing connections(?)")
      } yield ()
    } >> step("client exited").as(ExitCode.Success)
}
``` 
Run with `-Djdk.httpclient.HttpClient.log=trace` and `-Djdk.httpclient.keepalive.timeout=10`. (Scala CLI command: `scala-cli run --java-opt -Djdk.httpclient.HttpClient.log=trace --java-opt -Djdk.httpclient.keepalive.timeout=10 -S 2.13.16 repro.scala --`)

Run without leaking:
```
% scala-cli run --java-opt -Djdk.httpclient.HttpClient.log=trace --java-opt -Djdk.httpclient.keepalive.timeout=10 -S 2.13.16 repro.scala --
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
38438
---------- starting non-leaking code ----------
Jun 03, 2025 1:18:48 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: Applying request filters
Jun 03, 2025 1:18:48 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: Applying jdk.internal.net.http.AuthenticationFilter@664d031e
Jun 03, 2025 1:18:48 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: Applying jdk.internal.net.http.RedirectFilter@3df72b50
Jun 03, 2025 1:18:48 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: All filters applied
Jun 03, 2025 1:18:48 PM jdk.internal.net.http.HttpClientImpl registerTimer
INFO: MISC: Registering timer ConnectTimerEvent, TimeoutEvent[id=1, duration=PT10S, deadline=2025-06-03T13:18:58.464610080Z]
Jun 03, 2025 1:18:48 PM jdk.internal.net.http.HttpClientImpl cancelTimer
INFO: MISC: Canceling timer ConnectTimerEvent, TimeoutEvent[id=1, duration=PT10S, deadline=2025-06-03T13:18:58.464610080Z]
Jun 03, 2025 1:18:51 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: Applying response filters
Jun 03, 2025 1:18:51 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: Applying jdk.internal.net.http.RedirectFilter@3df72b50
Jun 03, 2025 1:18:51 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: Applying jdk.internal.net.http.AuthenticationFilter@664d031e
Jun 03, 2025 1:18:51 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: All filters applied
Jun 03, 2025 1:18:51 PM jdk.internal.net.http.Http1Response onFinished
INFO: MISC: Attempting to return connection to the pool: PlainHttpConnection: HttpConnection: java.nio.channels.SocketChannel[connected local=/127.0.0.1:41830 remote=localhost/127.0.0.1:8081]
Jun 03, 2025 1:18:51 PM jdk.internal.net.http.HttpConnection closeOrReturnToCache
INFO: MISC: Returning connection to the pool: PlainHttpConnection: HttpConnection: java.nio.channels.SocketChannel[connected local=/127.0.0.1:41830 remote=localhost/127.0.0.1:8081]
---------- non-leaking code exited ----------
---------- keeping alive for 15 seconds (connections will close after 10 s) ----------
Jun 03, 2025 1:19:01 PM jdk.internal.net.http.PlainHttpConnection close
INFO: MISC: Closing: PlainHttpConnection: HttpConnection: java.nio.channels.SocketChannel[connected local=/127.0.0.1:41830 remote=localhost/127.0.0.1:8081]
Jun 03, 2025 1:19:01 PM jdk.internal.net.http.HttpClientImpl cancelTimer
INFO: MISC: Canceling timer ConnectTimerEvent, TimeoutEvent[id=1, duration=PT10S, deadline=2025-06-03T13:18:58.464610080Z]
---------- closing connections(?) ----------
Jun 03, 2025 1:19:06 PM jdk.internal.net.http.HttpClientImpl$SelectorManager run
INFO: MISC: HttpClient-1-SelectorManager: HttpClient finished. Exiting...
Jun 03, 2025 1:19:06 PM jdk.internal.net.http.HttpClientImpl$SelectorManager shutdown
INFO: MISC: HttpClient-1-SelectorManager: shutting down
---------- client exited ----------
```

Run with leaking:
```
% scala-cli run --java-opt -Djdk.httpclient.HttpClient.log=trace --java-opt -Djdk.httpclient.keepalive.timeout=10 -S 2.13.16 repro.scala -- leak
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
38974
---------- starting leaking code ----------
Jun 03, 2025 1:24:29 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: Applying request filters
Jun 03, 2025 1:24:29 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: Applying jdk.internal.net.http.AuthenticationFilter@d1f2aed
Jun 03, 2025 1:24:29 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: Applying jdk.internal.net.http.RedirectFilter@12e2e860
Jun 03, 2025 1:24:29 PM jdk.internal.net.http.MultiExchange requestFilters
INFO: MISC: All filters applied
Jun 03, 2025 1:24:29 PM jdk.internal.net.http.HttpClientImpl registerTimer
INFO: MISC: Registering timer ConnectTimerEvent, TimeoutEvent[id=1, duration=PT10S, deadline=2025-06-03T13:24:39.919649937Z]
Jun 03, 2025 1:24:29 PM jdk.internal.net.http.HttpClientImpl cancelTimer
INFO: MISC: Canceling timer ConnectTimerEvent, TimeoutEvent[id=1, duration=PT10S, deadline=2025-06-03T13:24:39.919649937Z]
timed out
---------- leaking code exited ----------
---------- keeping alive for 15 seconds (connections will close after 10 s) ----------
Jun 03, 2025 1:24:33 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: Applying response filters
Jun 03, 2025 1:24:33 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: Applying jdk.internal.net.http.RedirectFilter@12e2e860
Jun 03, 2025 1:24:33 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: Applying jdk.internal.net.http.AuthenticationFilter@d1f2aed
Jun 03, 2025 1:24:33 PM jdk.internal.net.http.MultiExchange responseFilters
INFO: MISC: All filters applied
---------- closing connections(?) ----------
# the process hangs here and never exits
```

```
# In a separate terminal
% lsof -p 38974
...
java    38974 rasmus.precenth  93u     IPv6             103576       0t0     TCP localhost:53658->localhost:sunproxyadmin (CLOSE_WAIT)
```
[fiberdump.txt](https://github.com/user-attachments/files/20574713/fiberdump.txt)
[threaddump.txt](https://github.com/user-attachments/files/20574714/threaddump.txt)

In short, cancelling the requests leaves the connection in a stuck state that it never recovers from.

I can see that the code tries its best to juggle cancelable and uncancelable effects, but that seems to not be enough in this case.

vladislavsheludchenkov:
Closes #1188

The original issue this PR was aiming to solve was fixed in cats-effect, but only when running Java 16 and above. Cancellation is not supported below Java 16, and attempting to do so still leaks connections, so this PR makes the requests uncancelable instead of silently leaking them.

Original PR description before cats-effect fix:

Turns out cancellation of a completable future is different here based on `mayInterruptIfRunning` argument: if false, it reports successful cancellation but allows the underlying request to continue and leak resources since we aren't going to subscribe to its body; if true, it actually cancels the underlying request

It's not documented anywhere publicly, but mentioned in internal javadocs:

https://github.com/openjdk/jdk21/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/MultiExchange.java#L272

https://github.com/openjdk/jdk21/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/common/Cancelable.java#L38

`CompletableFutureTerminationTest` has been modified to fail if cancelled using `mayInterruptIfRunning` equal to `false`

Since the value for `mayInterruptIfRunning` is hardcoded in cats-effect, I had to copy the whole `fromCompletableFuture`. This isn't intended to be a permanent solution, my next steps will be creating a pull request to cats-effect to allow parametrization in that method and then using it in this repo once it's released, but opening this PR first to (1) get your feedback on if it's the right solution and (2) fix it as soon as possible since it looks like a major issue. Of course, let me know if you don't think it's the correct course of action and you prefer me starting with a PR to cats-effect

### DIFF
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala
@@ -178,7 +178,10 @@ object JdkHttpClient {
     ): Resource[F, Response[F]] =
       Resource
         .makeFull { (poll: Poll[F]) =>
-          (Deferred[F, Unit], poll(responseF)).tupled
+          val maybeCancelableResponseF =
+            if (JdkVersion.supportsCancellation) poll(responseF)
+            else responseF
+          (Deferred[F, Unit], maybeCancelableResponseF).tupled
         } { case (subscription, response) =>
           subscription.tryGet.flatMap {
             case None =>
@@ -274,7 +277,7 @@ object JdkHttpClient {
       F.delay {
         val builder = HttpClient.newBuilder()
         // workaround for https://github.com/http4s/http4s-jdk-http-client/issues/200
-        if (Runtime.version().feature() == 11) {
+        if (JdkVersion.tls13TriggersDeadlock) {
           val params = javax.net.ssl.SSLContext.getDefault().getDefaultSSLParameters()
           params.setProtocols(params.getProtocols().filter(_ != "TLSv1.3"))
           val _ = builder.sslParameters(params)
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkVersion.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkVersion.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2019 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.jdkhttpclient
+
+private object JdkVersion {
+  val supportsCancellation: Boolean = Runtime.version().feature() >= 16
+
+  // see https://github.com/http4s/http4s-jdk-http-client/issues/200
+  val tls13TriggersDeadlock: Boolean = Runtime.version().feature() == 11
+}
--- a/core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
+++ b/core/src/main/scala/org/http4s/jdkhttpclient/JdkWSClient.scala
@@ -104,11 +104,12 @@ object JdkWSClient {
                   handleReceive(error.asLeft); ()
                 }
               }
-              webSocket <- poll(
-                F.fromCompletableFuture(
-                  F.delay(wsBuilder.buildAsync(URI.create(req.uri.renderString), wsListener))
-                )
+              webSocketF = F.fromCompletableFuture(
+                F.delay(wsBuilder.buildAsync(URI.create(req.uri.renderString), wsListener))
               )
+              webSocket <-
+                if (JdkVersion.supportsCancellation) poll(webSocketF)
+                else webSocketF
               sendSem <- Semaphore[F](1L)
             } yield (webSocket, queue, closedDef, sendSem)
           } { case (webSocket, queue, _, _) =>
--- a/core/src/test/scala/org/http4s/jdkhttpclient/CompletableFutureTerminationTest.scala
+++ b/core/src/test/scala/org/http4s/jdkhttpclient/CompletableFutureTerminationTest.scala
@@ -36,6 +36,7 @@ import java.util.concurrent.TimeUnit
 import scala.concurrent.duration._
 
 final class CompletableFutureTerminationTest extends CatsEffectSuite {
+
   import CompletableFutureTerminationTest._
 
   private val duration: FiniteDuration =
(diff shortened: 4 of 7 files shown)


armanbilge:
Thanks! This was already fixed in Cats Effect, I think we just need to release.
- https://github.com/typelevel/cats-effect/pull/4375

vladislavsheludchenkov:
Perfect! Was looking at series/3.x branch and missed this change. Do you know if a new patch version of cats effect is getting released anytime soon? If yes, feel free to close this PR. Having a test that can detect this leak is probably still useful, but I can open a new PR after the bump

Prillan:
@armanbilge Would it be possible to have a 3.6.2 release for Cats Effect soon to fix this? We have an ugly workaround to prevent cancellation of requests in the most impacted places, but you never know when it's going to start leaking connections elsewhere

hamnis:
If the 3.6.2 is far off, i am not against merging this.

vladislavsheludchenkov:
Right, the cancellation test is failing on Java 11. That makes sense, as cancellation was implemented in version 16. Not sure how we should handle this. Branch in runtime on java version and make our effect uncancellable if it’s lower than 16?

hamnis:
I think we have some JDK version checking here somewhere else, so that would not be unreasonable.

https://github.com/http4s/http4s-jdk-http-client/blob/series/0.10/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala#L277

vladislavsheludchenkov:
Rebased and updated this PR to specifically target the issue with JDK < 16, since it's already fixed now when running JDK >= 16

